PR #14, opened 2017-12-12
Added support for RPD and RPIQ

pierreroudier:
Hi,

Hopefully this is of interest and within scope of `yardstick`,

I've added methods to compute the RPD and RPIQ  --- respectively the Ratio of Performance to Deviation (Williams, 1987) and Ratio of Performance to Inter-Quartile (Bellon-Maurel et al., 2010) --- which are ubiquitous metrics in two of my research fields (spectroscopy and digital soil mapping). While RPD is similar to R-squared, RPIQ has the advantage to be suitable when data is not normally distributed.


R/regression.R
```diff
@@ -179,20 +191,75 @@ ccc.data.frame <-
...
+        truth = !!enquo(truth),
+        estimate = !!enquo(estimate),
+        ...
+      )
+    data <- data[, c(vars$truth, vars$estimate)]
+    if (na.rm)
+      data <- data[complete.cases(data), ]
+
+    rmse <- rmse_calc(data[[vars$truth]], data[[vars$estimate]])
+    qs <- quantile(data[[vars$truth]], probs = seq(0, 1, 0.25), names = FALSE)
```
topepo: Perhaps just use `IQR` here instead of these two lines

topepo:
I'd never heard of these but they seem like good additions. Thanks!

Can you:

 * break them out into their own file? They're a little different than the standard metrics and should probably be in another file.
 * create unit tests based on some published data (or by doing the calculations "by hand" in the test code?
 * write a little bit about what they are and where they are applicable.
* isolate the actual computations in separate functions `rpd_calc` and `rpiq_calc` (as is done for `rmse` and others). I think that we should also have a non-data-frame interface for the regression metrics.

pierreroudier:
@topepo Thanks a lot for the feedback! Any file name you would like me to use in particular?

topepo:
That all looks good! Thanks

PR #18, opened 2018-05-23
add survival metrics

csetraynor:
Hi, I haven't been able to import a typeof "closure" in particular:

```
> library(prodlim)
> typeof(Hist)
[1] "closure"
```
which is needed to calculate the ibrier that is why I included survival in dependencies.


DavisVaughan:
I really appreciate the effort here, but I don't think these currently fit the rest of the yardstick API. We will keep these in mind for yardstick going forward, and will probably circle back to survival metrics at some point. Thanks

Issue #29, opened 2018-08-29
A bug in mnLogLoss


grayskripko:
``` r
library(yardstick)
#> Loading required package: broom
x <- structure(list(
    No = c(0.860384856004899, 1, 1), 
    Yes = c(0.139615143995101, 0, 0), 
    prob = c(0.139615143995101, 0, 0), 
    estimate = structure(c(1L, 1L, 1L), .Label = c("No", "Yes"), class = "factor"), 
    truth = structure(c(2L, 1L, 2L), .Label = c("No", "Yes"), class = "factor"), 
    truth_num = c(1, 0, 1)), 
    row.names = c(NA, -3L), class = c("tbl_df", "tbl", "data.frame"))
print(x)
#> # A tibble: 3 x 6
#>      No   Yes  prob estimate truth truth_num
#>   <dbl> <dbl> <dbl> <fct>    <fct>     <dbl>
#> 1 0.860 0.140 0.140 No       Yes           1
#> 2 1     0     0     No       No            0
#> 3 1     0     0     No       Yes           1
mnLogLoss(x[1:2,], truth = truth, No, Yes)
#> [1] -0.9844328
mnLogLoss(x, truth = truth, No, Yes)
#> Error in sum(apply(res, 1, pos_log)): invalid 'type' (list) of argument
```

Bug location https://github.com/tidymodels/yardstick/blob/master/R/probs.R#L191

topepo:
I'll merge this branch in after another PR is finished. Note below the name change and the output type is different (you can use the `_vec` version of the function to get the scalar back).

```r
> library(yardstick)
For binary classification, the first factor level is assumed to be the event.
Set the global option `yardstick.event_first` to `FALSE` to change this.

> #> Loading required package: broom
> x <- structure(list(
+     No = c(0.860384856004899, 1, 1), 
+     Yes = c(0.139615143995101, 0, 0), 
+     prob = c(0.139615143995101, 0, 0), 
+     estimate = structure(c(1L, 1L, 1L), .Label = c("No", "Yes"), class = "factor"), 
+     truth = structure(c(2L, 1L, 2L), .Label = c("No", "Yes"), class = "factor"), 
+     truth_num = c(1, 0, 1)), 
+     row.names = c(NA, -3L), class = c("tbl_df", "tbl", "data.frame"))
> print(x)
# A tibble: 3 x 6
     No   Yes  prob estimate truth truth_num
[...]

PR #37, opened 2018-10-05
multinomial log loss changes (bug fix and name change)

topepo:
@DavisVaughan and @ClaytonJY note the name change.


topepo:
closes #29

PR #64, opened 2018-11-22
Adding Gini coefficient metric

dandermotj:
Hi, 

I hope you don't mind this addition. Adding Gini as a metric might seem needless given it so simple to compute with the AUC, but I think there are two important reasons for its inclusion. Firstly the Gini is practically the only model performance metric used in financial risk analytics (particularly for credit scorecards) and secondly, many analysts/practitioners aren't aware of the connection between AUC and Gini. 

So by including it, yardstick becomes an option for analysts who don't know any better and many more people will find yardstick when they search for "compute gini scorecard" online. 

`gini` simply uses `2 * roc_auc_estimator_impl(...) - 1` as it's internal implementation and essentially the rest of the code is a copy/paste of the `roc_auc` function. I've also added three tests in `test-prob-roc.R`, which are also copies of the `roc_auc` tests, just with expected result scaled for the Gini. 

Let me know if I need to make any additions or changes for this to be accepted.

Thanks,
Dan


DavisVaughan:
I'm fine with this. We have both recall and sensitivity so I don't mind partial repetitiveness. I'll look at the PR in the next few days and get back to you!

dandermotj:
Great, thanks Davis

topepo:
Thanks for adding this. 

We tend to only document the author at the package level. You can remove that from the `gini` documentation.

dandermotj:
I've removed the the author attribution from the documentation. In case you're wondering, the forced push was for an amended commit - I had forgotten to rebuild the docs the first time.

DavisVaughan:
Hi @dandermotj, believe it or not, we actually already have this metric under the name `gain_capture()`. Let me try and explain.

You are proposing the Gini coefficient. There is an identical metric (as you mentioned in the docs) called AR for accuracy ratio.

`gini = accuracy_ratio`

The accuracy ratio is closely related to the curve that it goes with, the CAP Curve (cumulative accuracy profile). https://en.wikipedia.org/wiki/Cumulative_accuracy_profile

To do the AR calculation without the shortcut formula that uses AUC, you'd find:

> the ratio of the area between the model CAP and the random CAP and the area between the perfect CAP and the random CAP.    -from the Wiki

Okay, sounds good. Now, there is this other curve called the Gain Curve that we have `gain_curve()` for.
https://stats.stackexchange.com/questions/279385/gain-curve-interpretation

If you notice, it looks a lot like the CAP Curve. In fact they are the same! Both measure "the cumulative number of samples with true results relative to the entire number of true results." -me in the gain curve docs.

Okay, so if you can accept that `cap curve = gain curve`, then go look at the `?gain_capture` docs.

> gain_capture() calculates the area under the gain curve, but above the baseline, and then divides that by the area under a perfect gain curve, but above the baseline. It is meant to represent the amount of potential gain "captured" by the model.

This is exactly what `gini`/`AR` does!

As some proof, here is the results of `gini()` vs `gain_capture()`. (I've had to modify your `gini` function a bit as the generalization to the multiclass case had a mistake in it, it's a tough thing to get right. I've gone ahead and pushed to your PR so you can see the fix).

``` r
suppressPackageStartupMessages({
  library(yardstick)
  library(dplyr)
})

gini(two_class_example, truth, Class1)
#> # A tibble: 1 x 3
#>   .metric .estimator .estimate
#>   <chr>   <chr>          <dbl>
#> 1 gini    binary         0.879
gain_capture(two_class_example, truth, Class1)
#> # A tibble: 1 x 3
#>   .metric      .estimator .estimate
#>   <chr>        <chr>          <dbl>
#> 1 gain_capture binary         0.879

hpc_cv %>%
  group_by(Resample) %>%
  gini(obs, VF:L)
#> # A tibble: 10 x 4
#>    Resample .metric .estimator .estimate
#>    <chr>    <chr>   <chr>          <dbl>
#>  1 Fold01   gini    macro          0.743
#>  2 Fold02   gini    macro          0.727
#>  3 Fold03   gini    macro          0.796
#>  4 Fold04   gini    macro          0.748
#>  5 Fold05   gini    macro          0.730
#>  6 Fold06   gini    macro          0.754
#>  7 Fold07   gini    macro          0.730
#>  8 Fold08   gini    macro          0.747
#>  9 Fold09   gini    macro          0.710
#> 10 Fold10   gini    macro          0.731

hpc_cv %>%
  group_by(Resample) %>%
  gain_capture(obs, VF:L)
#> # A tibble: 10 x 4
#>    Resample .metric      .estimator .estimate
#>    <chr>    <chr>        <chr>          <dbl>
#>  1 Fold01   gain_capture macro          0.743
#>  2 Fold02   gain_capture macro          0.727
#>  3 Fold03   gain_capture macro          0.796
#>  4 Fold04   gain_capture macro          0.748
#>  5 Fold05   gain_capture macro          0.730
#>  6 Fold06   gain_capture macro          0.754
#>  7 Fold07   gain_capture macro          0.730
#>  8 Fold08   gain_capture macro          0.747
#>  9 Fold09   gain_capture macro          0.710
#> 10 Fold10   gain_capture macro          0.731
```

DavisVaughan:
So, after all that explanation, I think it would be best to add notes to `gain_curve()` and `gain_capture()`. 

For `gain_curve()` I will mention that it is identical to a CAP curve

For `gain_capture()` I will mention that it is identical to the gini coefficient and to the AR statistic.

For both I will include the link to the new paper that you included.

So this is a good thing! Thanks for your initial work on this.

dandermotj:
Hi Davis,

Thanks for your detailed response. I'd never even heard of the gain curve - you learn something new everyday! Also, thanks for adding the bug fix commit for the multi-class commit - if only to see what the internal implementation should have been.  

I can add to the group of metrics that seem different but are in fact equivalent to AR/Gini/Gain Capture. For binary outcomes, Somer's D is also equivalent to this group - you'll see a version of the AUC*2 - 1 formula in the Somer's D Wikipedia page. I don't know how this extends to the multi-class case, but I think given it's a measure of ordinal association I'd guess it's different. I saw it was listed in an issue as to be added, so might be something to note. Tidymodels is great, keep up the good work!

Cheers,
Dan

PR #74, opened 2019-01-19
Add Pseudo-Huber loss

### LINKED ISSUE #71: Huber Loss
This is a numeric metric.

I like this metric as a mix of MSE and MAE. Apparently used in robust regression? 
https://en.wikipedia.org/wiki/Huber_loss

It looks like the implementation would be:

equation

* I would use `mae()` as a guide. 

- [ ] Call it `huber_loss()`

- [ ] It would require a `delta` argument for the cutoff value. The standard seems to be to default this to `1`. See `ccc()` to learn how to use optional arguments.

- [ ] Pay close attention to how we generate documentation and examples automatically

- [ ] Try and understand how `metric_summarizer()` works, along with the rationale behind `metric_vec_template()`, then use them in the implementation.

- [ ] Add a few tests. Small examples you can match from academic papers are best, but otherwise any online example is okay. If all else fails, create an example "by hand" that is easy to manually compute the result for.

- [ ] Use the file naming scheme `num-<metric>.R`

If you take on this metric, I would also encourage you to implement the Pseudo-Huber loss function if you have time (maybe do a separate Pull Request for it). It is on the same page as the huber loss, just scroll down. I would call it `huber_loss_pseudo()`.

- [ ] In the `seealso` section of the docs, cross link to `huber_loss()`

Tensorflow defines this metric:
https://www.tensorflow.org/api_docs/python/tf/losses/huber_loss

The [Custom Metrics](https://tidymodels.github.io/yardstick/articles/custom-metrics.html) vignette will probably be helpful.

blairj09:
fixes #71 

- I did not provide an additional reference. One may be necessary, but I couldn't find an explicit reference to the [Wikipedia implementation](https://en.wikipedia.org/wiki/Huber_loss#Pseudo-Huber_loss_function)


DavisVaughan:
Found a reference to it on page 619 of http://cvrs.whu.edu.cn/downloads/ebooks/Multiple%20View%20Geometry%20in%20Computer%20Vision%20(Second%20Edition).pdf

DavisVaughan:
Thanks!

Issue #67, opened 2019-01-05
Symmetric MAPE


DavisVaughan:
Definition:
https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error

(Use the definition with absolute values in the denominator)

I would use `mape()` as a guide. 

- [ ] Pay close attention to how we generate documentation and examples automatically

- [ ] Try and understand how `metric_summarizer()` works, along with the rationale behind `metric_vec_template()`, then use them in the implementation.

- [ ] Add a few tests. Small examples you can match from academic papers are best, but otherwise any online example is okay. If all else fails, create an example "by hand" that is easy to manually compute the result for.

- [ ] Use the file naming scheme `num-<metric>.R`

The [Custom Metrics](https://tidymodels.github.io/yardstick/articles/custom-metrics.html) vignette will probably be helpful.

PR #76, opened 2019-01-21
add symmetric mape to yardstick

riazhedayati:
fixes issue #67


topepo:
I think that you need to run the documentation generation; the function has an export tag but the NAMESPACE file has not been modified. You can do this by opening the `yardstick.Rproj` file and use the "Document" button

image

then push the commit.

riazhedayati:
yup, I definitely forgot that step...thanks Max.

topepo:
Thanks

PR #79, opened 2019-01-30
missing validation in metrics

EmilHvitfeldt:
I noticed that `metrics()` was the only function that didn't validate the `truth` and `estimate` argument, so I added tests using the helper function `validate_not_missing()`.


DavisVaughan:
Thanks! Could you move these checks into `all_select()`? It's only currently called from `metrics()`, but if we ever use it anywhere else it might be useful to have those checks automatically.

EmilHvitfeldt:
Yes! The checks have been moved.

DavisVaughan:
Thanks!

Issue #10, opened 2017-11-09
confusion matrix plot function


topepo:
See topepo/caret#755

jminnier:
I'm going to give this a try!

PR #80, opened 2019-01-31
adding autoplot method for conf_mat

EmilHvitfeldt:
I gave a shot at #10, I couldn't decide between the standardly used heatmap and the mosaic plot so I added both.

``` r
library(tidymodels)
cm_small <- conf_mat(two_class_example, truth, predicted)

cm_big <- hpc_cv %>%
  filter(Resample == "Fold01") %>%
  conf_mat(obs, pred)

autoplot(cm_small)
```

``` r
autoplot(cm_big)
```

``` r

autoplot(cm_small, type = "heatmap")
```

``` r
autoplot(cm_big, type = "heatmap")
```

<sup>Created on 2019-01-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

with the standard ggplot-ness scales can be changed again

``` r
autoplot(cm_small, type = "heatmap") +
  scale_fill_gradient(low = "white", high = "darkred")

```

<sup>Created on 2019-01-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,9 @@
 * `smape()` is a numeric metric that is based on percentage errors 
 (@riazhedayati, #67).
 
+* `conf_mat()` got two `ggplot2::autoplot()` methods for easy visualization 
+(@EmilHvitfeldt, $10).
+
 # yardstick 0.0.2
 
 ## Breaking changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@
 (@riazhedayati, #67).
 
 * `conf_mat()` got two `ggplot2::autoplot()` methods for easy visualization 
-(@EmilHvitfeldt, $10).
+(@EmilHvitfeldt, #10).
 
 # yardstick 0.0.2
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,8 +10,8 @@
 * `smape()` is a numeric metric that is based on percentage errors 
 (@riazhedayati, #67).
 
-* `conf_mat()` got two `ggplot2::autoplot()` methods for easy visualization 
-(@EmilHvitfeldt, #10).
+* `conf_mat` objects now have two  `ggplot2::autoplot()`methods for easy visualization
+of the confusion matrix as either a heat map or a mosaic plot (@EmilHvitfeldt, #10).
 
 # yardstick 0.0.2
 
--- a/R/conf_mat.R
+++ b/R/conf_mat.R
@@ -83,7 +83,7 @@
 #'
 #' round(mean_cmat, 3)
 #'
-#' # The confusion matrix can quickly be vizualized using autoplot()
+#' # The confusion matrix can quickly be visualized using autoplot()
 #' library(ggplot2)
 #'
 #' autoplot(cm, type = "mosaic")
--- a/R/conf_mat.R
+++ b/R/conf_mat.R
@@ -9,6 +9,10 @@
 #' There is also a `summary()` method that computes various classification
 #' metrics at once. See [summary.conf_mat()]
 #'
+#' There is a [ggplot2::autoplot()]
+#' method for quickly visualizing the matrix. Both a heatmap and mosaic type
+#' is implemented.
+#'
 #' The function requires that the factors have exactly the same levels.
 #'
 #' @aliases conf_mat.table conf_mat.default conf_mat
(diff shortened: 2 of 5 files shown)


R/conf_mat.R
```diff
@@ -274,3 +286,90 @@ summary.conf_mat <- function(object,
...
+         heatmap = cm_heat(object))
+}
+
+
+
+cm_heat <- function(x) {
+
+  `%+%` <- ggplot2::`%+%`
+
+  data <- as.data.frame.matrix(x$table)
```
DavisVaughan: Do you need `as.data.frame.matrix`? `x$table` is a `table` and `as.data.frame.table` (the default method here) seems to do what you try and do manually with the rownames and `gather()`

R/conf_mat.R
```diff
@@ -77,6 +82,13 @@
 #' colnames(mean_cmat) <- levels(hpc_cv$obs)
 #'
 #' round(mean_cmat, 3)
+#'
+#' # The confusion matrix can quickly be vizualized using autoplot()
```
DavisVaughan: visualized*

DavisVaughan:
Can you please add a bullet to NEWS?

DavisVaughan:
What if the guts of `cm_heat` were more like:

```r
data %>% tidyr::gather(key = "Truth", value = "Count", -Prediction) %>% 
  ggplot2::ggplot(ggplot2::aes(Prediction, Truth, fill = Count)) %+% 
  ggplot2::geom_tile() %+% ggplot2::scale_fill_gradient(low = "grey90", 
                                                        high = "grey40") +
  ggplot2::theme(panel.background = ggplot2::element_blank(), legend.position = "none") +
  geom_text(aes(label = Count)) 
```

screen shot 2019-02-01 at 3 34 46 pm

I like that we don't have a white spot where the value is `0` because that blends with the background too much. I also like the label.

DavisVaughan:
The labels on the x-axis should match up with the order of the y-axis for the heat map (`VF` should be in the top left on the y-axis and the bottom left of the x-axis). That way the diagonal is the darkest section.

R/conf_mat.R
EmilHvitfeldt: you are right, `as.data.frame.table` does exactly what is needed

R/conf_mat.R
```diff
@@ -274,3 +286,90 @@ summary.conf_mat <- function(object,
...
+  data <- as.data.frame.matrix(x$table)
+
+  data$Prediction <- rownames(data)
+
+  data %>%
+    tidyr::gather(key = "Truth", value = "Count", -Prediction) %>%
+    ggplot2::ggplot(ggplot2::aes(Prediction, Truth, fill = Count)) %+%
+    ggplot2::geom_tile() %+%
+    ggplot2::scale_fill_gradient(low = "white", high = "black") %+%
+    ggplot2::theme_bw()
```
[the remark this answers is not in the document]
EmilHvitfeldt: That would be fine! fixing right now

EmilHvitfeldt:
The new additions (Thanks @DavisVaughan!) now give the following heatmaps

``` r
library(tidymodels)

cm_small <- conf_mat(two_class_example, truth, predicted)
cm_big <- hpc_cv %>%
  filter(Resample == "Fold01") %>%
  conf_mat(obs, pred)

autoplot(cm_small, type = "heatmap")
```



``` r
autoplot(cm_big, type = "heatmap")
```



<sup>Created on 2019-02-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/conf_mat.R
```diff
@@ -313,11 +313,14 @@ cm_heat <- function(x) {
   `%+%` <- ggplot2::`%+%`
 
   as.data.frame.table(x$table) %>%
-    ggplot2::ggplot(ggplot2::aes(Prediction, Truth, fill = Freq)) %+%
+    ggplot2::ggplot(ggplot2::aes(x = Prediction,
```
DavisVaughan: Would it be slightly better to do:

```
dplyr::mutate(Truth = factor(Truth, levels = rev(levels(Truth))) %>%
```

right before the ggplot2 call so then you don't need to update the label at the end?

NEWS.md
```diff
@@ -10,6 +10,9 @@
 * `smape()` is a numeric metric that is based on percentage errors 
 (@riazhedayati, #67).
 
+* `conf_mat()` got two `ggplot2::autoplot()` methods for easy visualization 
+(@EmilHvitfeldt, $10).
```
DavisVaughan: `$10`? did you mean `#80`? 🤣

DavisVaughan: I think I would say:

`conf_mat` objects now have

rather than

`conf_mat()` got

EmilHvitfeldt: Should I also reference #10?

DavisVaughan: oh yes please

DavisVaughan: reference only 10, not 80. sorry

EmilHvitfeldt: No problem, I was the one using $ instead of # 😄

DavisVaughan:
Thanks!

Issue #68, opened 2019-01-05
Mean Absolute Scaled Error (MASE)


DavisVaughan:
Definition:
https://en.wikipedia.org/wiki/Mean_absolute_scaled_error

See also the Scale Free Errors section of:
http://datascienceassn.org/sites/default/files/Another%20Look%20at%20Measures%20of%20Forecast%20Accuracy.pdf

- I would use `mape()` as a general guide.
- There is a seasonal and non-seasonal version. The non-seasonal version is just the seasonal version with `m = 1` so we can just implement 1 version that handles seasonality and defaults to the non-seasonal version.
- To see how extra arguments can be passed through, use `ccc()` as a guide. Specifically, you need to pass it through in the `metric_summarizer()` and in the metric_vec_template()`.

- [ ] Call it `mase()`

- [ ] Because of the seasonality, it will need a parameter to control that. I would call it `period = 1` and document clearly that this controls the seasonal period in the calculation. (this is `m` in the wiki documentation)

- [ ] Pay close attention to how we generate documentation and examples automatically

- [ ] Try and understand how `metric_summarizer()` works, along with the rationale behind `metric_vec_template()`, then use them in the implementation.

- [ ] Add a few tests. Small examples you can match from academic papers are best, but otherwise any online example is okay. If all else fails, create an example "by hand" that is easy to manually compute the result for.

- [ ] Use the file naming scheme `num-<metric>.R`

- [ ] I would use add a documentation reference to the paper by Hyndman, since it is a metric he helped develop http://datascienceassn.org/sites/default/files/Another%20Look%20at%20Measures%20of%20Forecast%20Accuracy.pdf

- [ ] In the docs, mention that this is a metric generally used in time series forecasting.

The [Custom Metrics](https://tidymodels.github.io/yardstick/articles/custom-metrics.html) vignette will probably be helpful.

ClaytonJY:
claimed!

alexhallam:
If it useful for testing here is the data from [Hyndman's paper](http://datascienceassn.org/sites/default/files/Another%20Look%20at%20Measures%20of%20Forecast%20Accuracy.pdf). 

```r
library(tidyverse)
sales <- tibble(g = c(rep("insample", 23), rep("outsample", 12)),
       y = c(0,2,0,1,0,10,0,0,0,2,0,6,3,0,0,0,0,0,7,0,0,0,0, 0, 0, 0, 3, 1, 0, 0, 1, 0, 1, 0, 0),
[...]

PR #81, opened 2019-02-01
mase error on insample

alexhallam:


alexhallam:
Not sure how to handle the insample argument

topepo:
Is this related to an issue?

alexhallam:
Yeah, I don't know how to properly add a new vector argument to the `mase` function. The issue is here #68

topepo:
So you want to compute the metric for different values of `insample`?

alexhallam:
No, the problem is that MASE needs to know what values are insample. It is a necessary part of the calculation.  MASE is the out-of-sample mae over the in sample mae of the seasonal naive forecast. For this calculation to be possible the user has to label what data is `insample`. 
```
mase <- out_mae / in_snaive_mae
```

DavisVaughan:
Ah, I see what you are trying to say. This is more difficult that I intended it to be. I did not read the definition of the metric correctly, and did not see that it used the in sample MAE as the denominator.

This is difficult because `yardstick` generally has no knowledge of the in sample outcome variable. It only gets the out of sample predictions, and the corresponding out of sample outcomes.

I have one suggestion, and I don't particularly love it, but it might be good enough.

`mase(data, truth, estimate, m = 1, mae_train = NULL)`

We could take `truth`, lag it by `m`, calculate `(truth - truth_lag)`, and compute the denominator using that seasonal 1 step naive forecast, rather than using a true in sample MAE. To me, this has the same rationale as using the in sample one.

If the user knew the in sample mae value, and really wanted to use that, they could pass that single numeric value to `mae_train` and that would override the calculation described above.

alexhallam:
Do you think that if `mae_train` is null that a warning should be thrown stating that this is an ad_hoc mase and that a value for `mae_train` is recommended? Or do you think this would be too forceful?

DavisVaughan:
Sounds too forceful since it's the default. I would definitely be clear in the documentation about the difference though!

PR #83, opened 2019-02-03
Working mase

alexhallam:


DavisVaughan:
I tweaked a few things in the documentation and implementation structure, added a bit more validation, and some tests to go along with it, but otherwise it is looking good! I will merge when travis is done.

DavisVaughan:
Thank you!

PR #110, opened 2019-06-01
Use tidy version of coords (xrobin/pROC#54)

xrobin:
An upcoming version of pROC (1.16) will feature backward incompatible changes in the coords function. In order to adhere to the principles of tidy datasets, the output will be transposed, and a data.frame returned instead of a matrix.

This PR sets the 'transpose' argument to 'coords' explicitly, and ensures proper operation with current (1.15) and future versions of pROC.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,7 @@ Suggests:
 Imports:
     dplyr,
     utils,
-    pROC,
+    pROC (>= 1.15.0),
     rlang,
     tidyselect,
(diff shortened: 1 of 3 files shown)


xrobin:
The build failed for some targets. It looks like it didn't refresh the build cache and used pROC 1.14.0 instead of 1.15.0. Maybe I opened this PR too quickly after the 1.15 release.

DavisVaughan:
This is great, thanks!

PR #115, opened 2019-08-11
Iic feature

### LINKED ISSUE #87: Add modified correlation statistic?
The [Ideality of Correlation statistic](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Ideality+of+Correlation%22&btnG=) is something that comp chemists seem to like (I guess RMSE isn't good enough for them). We might think about adding that. 

image

jyuu:
Fixes #87


R/num-iic.R
```diff
@@ -0,0 +1,88 @@
+#' Ideality of correlation statistic
```
DavisVaughan: Could you please change this to:

> Index of ideality of correlation

DavisVaughan: Actually I am going to make a few small tweaks so I can do it!

DavisVaughan:
Thanks @jyuu!

PR #125, opened 2019-10-18
Added kap_weighted for #2

jonthegeek:
I'll likely add more tests as I dive in, but I got the sklearn comparison set up.


jonthegeek:
This should work to finalize the first bullet for #2.

jonthegeek:
Hmm. The templated example fails because it's an (unordered) factor. Seeing if I can figure out how to fix that...

DavisVaughan:
You can use non templated examples for this one if its easier because of the orderedness

R/class-kap_weighted.R
```diff
@@ -0,0 +1,133 @@
...
+}
+
+kap_weighted_binary <- function(data, weight) {
+
+  n <- sum(data)
+  n_class <- dim(data)[[1]]
+  .row_sums <- rowSums(data)
+  .col_sums <- colSums(data)
+
+  weight_power <- switch(
```
DavisVaughan: I think you might be getting the `weight_power` twice? on line 112 and 123?

jonthegeek: Oops, yes! I solidified it with the switch but missed the spot where I'd originally simply defined it! Thanks for the catch. Updating in a sec...

jonthegeek:
@DavisVaughan I *finally* worked through all the Travis failures. Take a look and let me know what you think!

man-roxygen/examples-class-ordered.R
```diff
@@ -0,0 +1,43 @@
...
+#'
+#' # Groups are respected
+#' hpc_cv %>%
+#'   group_by(Resample) %>%
+#'   <%=metric_fn %>(obs, pred)
+#'
+#' # Weighted macro averaging
+#' hpc_cv %>%
+#'   group_by(Resample) %>%
+#'   <%=metric_fn %>(obs, pred, estimator = "macro_weighted")
```
DavisVaughan: If it extends naturally to multiclass, does this do anything?

tests/pycompare/generate-pycompare.R
```diff
@@ -48,3 +48,32 @@ py_kap <- list(
...
+truth <- as.integer(truth)
+estimate <- as.integer(estimate)
+
+py_kap_weighted_linear <- list(
+  binary = skmetrics$cohen_kappa_score(two_class_example$truth, two_class_example$predicted, weights = "linear"),
+  multiclass = skmetrics$cohen_kappa_score(truth, estimate, weights = "linear")
+)
+saveRDS(py_kap_weighted_linear,
+        "tests/pycompare/py-kap-weighted-linear",
+        version = 2)
```
DavisVaughan: Does the version matter?

jonthegeek: It mattered for Travis. The old-version-of-R check failed when I didn't include the version. It doesn't matter to *me*, per se, but I wanted to clear the red.

man-roxygen/examples-class-ordered.R
jonthegeek: I just copied from examples-class, just changing it to be ordered. I tried not to change too much for fear that it might blow up. Note that kap.Rd (which also extends naturally to multiclass) has the same extraneous "Weighted macro averaging" block.

PR #129, opened 2019-12-01
document that metrics() returns 3 column tibble

davechilders:
- `metrics()` appears to return a *three* column tibble with columns {`.metric`, .`estimator`, and .`estimate`}

### DIFF
--- a/R/metrics.R
+++ b/R/metrics.R
@@ -22,7 +22,7 @@
 #'
 #' @return
 #'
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Thanks! Next time, when you change documentation like this could you please also run `devtools::document()` to update the Rd files?

I've done it here, so need to do anything else.

PR #133, opened 2020-01-22
closes #130

### LINKED ISSUE #130: `conf_mat()` and it's `autoplot()` method ordering should match
The plot needs predictions along the `y` axis, and truth along the `x` axis

Screen Shot 2020-01-14 at 2 32 01 PM
Screen Shot 2020-01-14 at 2 32 05 PM

I think that for `cm_heat()` this isn't too bad. We can swap out for these lines:

```r
    dplyr::mutate(Prediction = factor(Prediction, levels = rev(levels(Prediction)))) %>%
    ggplot2::ggplot(ggplot2::aes(Truth, Prediction, fill = Freq)) %+%
```

For `cm_mosaic()` I think this gets us close, but the labels don't look like they are placed quite right

```r
  ggplot2::ggplot(full_data) %+%
    ggplot2::geom_rect(ggplot2::aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax)) %+%
    ggplot2::scale_x_reverse(
      breaks = (x_data$xmin + x_data$xmax) / 2,
      labels = tick_labels
    ) %+%
    ggplot2::scale_y_reverse(
      breaks = (y1_data$ymin + y1_data$ymax) / 2,
      labels = tick_labels
    ) %+%
    ggplot2::labs(x = "Predicted", y = "Truth") %+%
    ggplot2::theme(panel.background = ggplot2::element_blank()) %+%
    ggplot2::coord_flip()
```

topepo:
@apreshill

Maybe @EmilHvitfeldt could look at my changes to the mosaic plot to verify that this was the right approach.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -13,7 +13,7 @@ License: GPL-2
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 6.1.1
+RoxygenNote: 7.0.2.9000
 Depends:
     R (>= 2.10)
 Suggests: 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,9 @@
 
 ## Other improvements
 
- * The `autoplot()` methods for confusion matrices is now consistent with the `print()` method such that the true values are in columns. 
+* The `autoplot()` heat map for confusion matrices now places the predicted values on the `x` axis and the truth values on the `y` axis to be more consistent with the confusion matrix `print()` method.
+
+* The `autoplot()` mosaic plot for confusion matrices had the `x` and `y` axis labels backwards. This has been corrected.
 
 # yardstick 0.0.4
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # yardstick (development version)
 
+## Other improvements
+
+ * The `autoplot()` methods for confusion matrices is now consistent with the `print()` method such that the true values are in columns. 
+
 # yardstick 0.0.4
 
 ## New metrics and functionality
--- a/R/conf_mat.R
+++ b/R/conf_mat.R
@@ -347,7 +347,7 @@ cm_mosaic <- function(x) {
 
   `%+%` <- ggplot2::`%+%`
 
-  cm_zero <- (as.numeric(x$table == 0) / 2) + t(x$table)
+  cm_zero <- (as.numeric(x$table == 0) / 2) + x$table
 
   x_data <- space_fun(rowSums(cm_zero), 200)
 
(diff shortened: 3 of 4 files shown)


DavisVaughan:
Am I crazy, or is the original implementation correct, but the _labels_ are just on the wrong axes? If you swap where `Predicted` and `Truth` labels doesn't this look right? The proportions all look right to me

``` r
library(yardstick)
library(ggplot2)
library(dplyr)

data("hpc_cv")

cm <- hpc_cv %>%
  filter(Resample == "Fold01") %>%
  conf_mat(obs, pred)

cm
#>           Truth
#> Prediction  VF   F   M   L
#>         VF 166  33   8   1
#>         F   11  71  24   7
#>         M    0   3   5   3
#>         L    0   1   4  10

autoplot(cm, type = "mosaic")
```



<sup>Created on 2020-01-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

DavisVaughan:
@topepo my guess is that you would need to transpose `x$table` for all of its usages. i.e. with the current PR i get this, but a few things don't look right. The right column only has 3 bars. The first column, 3rd row is pretty wide even though it should have a 0 there.

``` r
library(yardstick)
#> For binary classification, the first factor level is assumed to be the event.
#> Set the global option `yardstick.event_first` to `FALSE` to change this.
library(ggplot2)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

data("hpc_cv")

cm <- hpc_cv %>%
  filter(Resample == "Fold01") %>%
  conf_mat(obs, pred)

cm
#>           Truth
#> Prediction  VF   F   M   L
#>         VF 166  33   8   1
#>         F   11  71  24   7
#>         M    0   3   5   3
#>         L    0   1   4  10

autoplot(cm, type = "mosaic")
```



<sup>Created on 2020-01-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

I imagine the other usage of `x$table` here would need to be transposed, but that gives the same plot as what I am proposing with just swapping the labels

https://github.com/tidymodels/yardstick/pull/133/files#diff-396c2efbb3e17c59af23275ff1380255R350

topepo:
So I think that the CRAN version does have the axis labels wrong for mosaic plots. That doesn't need different computations (i.e. transpose). 

The heatmap still needs to be transposed though. I'll update in a sec.

DavisVaughan:
I think it looks okay now, right?

Issue #132, opened 2020-01-19
roc_auc should return NA with no cases or controls


topepo:
``` r
library(tidymodels)
#> Registered S3 method overwritten by 'xts':
#>   method     from
#>   as.zoo.xts zoo
#> ── Attaching packages ─────────────────────────────────────────────────────────── tidymodels 0.0.4 ──
#> ✓ broom     0.5.2          ✓ recipes   0.1.9     
#> ✓ dials     0.0.4          ✓ rsample   0.0.5     
#> ✓ dplyr     0.8.3          ✓ tibble    2.1.3     
#> ✓ ggplot2   3.2.1          ✓ tune      0.0.1     
#> ✓ infer     0.5.1          ✓ workflows 0.1.0     
#> ✓ parsnip   0.0.4.9000     ✓ yardstick 0.0.4     
#> ✓ purrr     0.3.3
#> ── Conflicts ────────────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()
data(two_class_example)
two_class_example %>% 
  dplyr::filter(truth == "Class1") %>% 
  roc_auc(truth, Class1)
#> Error in roc.default(response, predictor, auc = TRUE, ...): No control observation.
```

<sup>Created on 2020-01-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It would be better to return `NA` with a warning.

DavisVaughan:
I think the main change to make is to `roc_auc_binary()`. It should have two checks:

- if `truth` doesn't have any `lvl[1]` values, then `NA_real_` should be returned with a warning that there were no events
- if `truth` doesn't have any of `lvl[2]` values, then `NA_real_` should be returned with a warning that there were no controls

At first glance it looks like this will be enough to fix macro/micro/macro_weighted multiclass cases too, since they use `roc_auc_binary()` internally

[...]

PR #138, opened 2020-02-01
fix roc_auc to return NA if no cases or controls

dpastling:
fix roc_auc_binary() to return NA if no cases or controls and generate
an informative warning

currently generates the following warning:

```
test_autoplot.R:80: warning: ROC Curve - multi class
Row indexes must be between 0 and the number of rows (0). Use `NA` as row index to obtain a row full of `NA` values.
This warning is displayed once per session.
```

issue #132

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # yardstick (development version)
 
+* `roc_auc()` now warns when there are no events or controls in the provided `truth` column, and returns `NA` (@dpastling, #132).
+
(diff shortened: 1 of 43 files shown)


DavisVaughan:
Thanks @dpastling! Also, that warning ended up coming from a combination of the development version of tibble + the cran version of ggplot2, it had nothing to do with anything you changed!

dpastling:
Thanks @DavisVaughan for cleaning up the code. I didn't have much time at the end to add additional tests. Glad to know the warning was from package versions and not because of NAs introduced elsewhere in the codebase.

Issue #158, opened 2020-05-20
autoplot(type = "mosaic") has incorrect widths


richierocks:
- [x] Start a new R session
- [x] Install the latest version of of the package: `update.packages(oldPkgs="yardstick", ask=FALSE)`
- [x] [Write a minimal reproducible example](http://stackoverflow.com/a/5963610)
- [x] run `sessionInfo()` and add the results to the issue. Even better would be to use the [`sessioninfo`](https://github.com/r-lib/sessioninfo) package's `session_info()`.  

### Minimal, reproducible example:

1. Create a confusion matrix.
2. Call `autoplot()` with the default `type = "mosaic"`.

#### Minimal, runnable code:

``` r
library(ggplot2)
library(yardstick)

actual_response <- rep(c(0, 1, 0, 1), 1:4)
predicted_response <- rep(c(0, 0, 1, 1), 1:4)
outcomes <- table(predicted_response, actual_response)
(confusion <- conf_mat(outcomes))
#>                   actual_response
#> predicted_response 0 1
#>                  0 1 2
#>                  1 3 4


autoplot(confusion)
```

![](https://i.imgur.com/nOBPDgu.png)

Since there are 4 actual responses of 0 and 6 actual responses of 1, I expected the widths of the columns to be a 40-60 split. They are actually a 30-70 split.

It looks like the problem is the use of `rowSums()` instead of `colSums()`, [here](https://github.com/tidymodels/yardstick/blob/master/R/conf_mat.R#L352).

For comparison, here's the equivalent plot using `vcd::mosaic()`.

``` r
library(vcd)
actual_response <- rep(c(0, 1, 0, 1), 1:4)
predicted_response <- rep(c(0, 0, 1, 1), 1:4)
mosaic(~ actual_response + predicted_response, direction = c("v", "h"))
```

![](https://i.imgur.com/36NIgjW.png)

You can see the column width split is different here.

### Session Info:

```
─ Session info ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value                       
 version  R version 4.0.0 (2020-04-24)
 os       macOS Mojave 10.14.6        
 system   x86_64, darwin17.0          
 ui       RStudio                     
 language (EN)                        
 collate  en_US.UTF-8                 
 ctype    en_US.UTF-8                 
 tz       America/New_York            
 date     2020-05-20                  

─ Packages ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[...]

PR #159, opened 2020-05-21
Fix autoplot() mosaic plot column widths

richierocks:
Fixes https://github.com/tidymodels/yardstick/issues/158


richierocks:
The test coverage check is failing on the "Install dependencies" step, but I can't see how that would be related to my patch.

DavisVaughan:
Thanks!

PR #164, opened 2020-06-11
fix typo

trang1618:


juliasilge:
Thanks for this PR @trang1618! Can you describe a little more why you think the curly brackets might help? Without the brackets, the documentation for F Measure looks like this:

```
F_meas_β = (1+β^2) * precision * recall/((β^2 * precision)+recall)
```

With the brackets, it renders in the `.Rd` file like this:

```
F_{meas}_β = (1+β^2) * precision * recall/((β^2 * precision)+recall)
```

I am not finding it clearer or easier to parse with the brackets. Was there something else you were thinking or intending?

trang1618:
Thanks for the quick feedback @juliasilge! Without the curly braces, this is what I see on the yardstick page for `f_meas`, which I don't think is what the equation is intended to look like.
https://yardstick.tidymodels.org/reference/f_meas.html


f_meas_eqn

juliasilge:
Ah, I was looking at how the docs render in `.Rd` files, not the pkgdown site. Makes sense!

@DavisVaughan the underscores used here are making things render really differently in `.Rd` and pkgdown. [Wikipedia just does `F_beta`](https://en.wikipedia.org/wiki/F1_score). What do you think about that?

DavisVaughan:
@juliasilge I think I am more inclined to do `F_meas`, without the beta part, since the metric is called `f_meas()` and you can see the beta in the RHS of the formula pretty clearly. Does that sound ok?

juliasilge:
@DavisVaughan if we do that, we'll need to use the curly brackets. It will render in `.Rd` like this:

```
F_{meas} = (1+β^2) * precision * recall/((β^2 * precision)+recall)
```

And the `meas` will render as subscript in pkgdown. Do you think that's clear/nice?

man-roxygen/table-relevance.R
```diff
@@ -9,6 +9,6 @@
 #'
 #' \deqn{recall = A/(A+C)}
 #' \deqn{precision = A/(A+B)}
-#' \deqn{F_meas_\beta = (1+\beta^2) * precision * recall/((\beta^2 * precision)+recall)}
+#' \deqn{F_{meas}_\beta = (1+\beta^2) * precision * recall/((\beta^2 * precision)+recall)}
```
juliasilge, line 12:
```suggestion
#' \deqn{F_{meas} = (1+\beta^2) * precision * recall/((\beta^2 * precision)+recall)}
```
Based on chatting with @DavisVaughan

juliasilge:
Thank you so much for bringing this to our attention @trang1618! 🚀

PR #165, opened 2020-06-16
add concordance statistic from survival package

EmilHvitfeldt:
This PR adds the concordance statistic from the **survival** package.
A couple of notes:

- I have named the function `c_index()` to avoid clashing with the concordance S3method from **survival**. I am still mentioning both C-Index and concordance in the documentation.
- **survival** has been added to Imports.


DavisVaughan:
Just an FYI that we are going to wait on this for a while longer as we figure out exactly what a survival model API looks like

Issue #160, opened 2020-05-21
Store event_first in conf_mat objects


richierocks:
Currently, for binary classification, it seems that whether or not the event is the first factor level can only be set globally, via `options(yardstick.event_first = FALSE)`.

Semantically, it feels as though the event level ought to be a property of the confusion matrix. By storing the event column in the `conf_mat` object, you can more easily work with a mix of response types where sometimes the event is the first level, and other times it is not.

### Minimal, reproducible example:

First, we define two confusion matrices, one where the event isn't the first value, and another where it is.

``` r
library(ggplot2)
library(yardstick)
#> For binary classification, the first factor level is assumed to be the event.
#> Set the global option `yardstick.event_first` to `FALSE` to change this.
library(withr)

actual_response1 <- rep(c(0, 1, 0, 1), 1:4)
predicted_response1 <- rep(c(0, 0, 1, 1), 1:4)
outcomes1 <- table(predicted_response1, actual_response1)
(confusion1 <- conf_mat(outcomes1))
#>                    actual_response1
#> predicted_response1 0 1
#>                   0 1 2
#>                   1 3 4

actual_response2 <- factor(rep(c("neg", "pos", "neg", "pos"), 1:4), levels = c("pos", "neg"))
predicted_response2 <- factor(rep(c("neg", "neg", "pos", "pos"), 1:4), levels = c("pos", "neg"))
outcomes2 <- table(predicted_response2, actual_response2)
(confusion2 <- conf_mat(outcomes2))
#>                    actual_response2
#> predicted_response2 pos neg
#>                 pos   4   3
#>                 neg   2   1
```

Now if you want to calculate performance metrics, you have to write grim code that specifies a global option.

```r
with_options(
  c(yardstick.event_first = FALSE),
  summary(confusion1)
)
#> # A tibble: 13 x 3
#>    .metric              .estimator .estimate
#>    <chr>                <chr>          <dbl>
#>  1 accuracy             binary        0.5   
#>  2 kap                  binary       -0.0870
#>  3 sens                 binary        0.667 
#>  4 spec                 binary        0.25  
#>  5 ppv                  binary        0.571 
#>  6 npv                  binary        0.333 
#>  7 mcc                  binary       -0.0891
#>  8 j_index              binary       -0.0833
#>  9 bal_accuracy         binary        0.458 
#> 10 detection_prevalence binary        0.7   
#> 11 precision            binary        0.571 
#> 12 recall               binary        0.667 
[...]

PR #172, opened 2020-07-08
Soft deprecate `yardstick.event_first` in favor of explicit `event_level` argument

### LINKED ISSUE #163: Consider soft deprecating the `yardstick.event_first` option
This is a global option that changes computation. This is generally a bad thing, so I'd much rather it be an argument to the metric function. Possibly we could push it into the `estimator` argument with `"binary"` continuing to do use the 1st level as the event and `"binary-last"` using the 2nd level.

This makes sense since it is ignored anyways for macro/micro multiclass settings.

This would help with #160 since we could then provide that option to `conf_mat()` and pass it through to all the metrics that `summary.conf_mat()` computes. Or make it an option to `summary.conf_mat()`

We could warn if the user had set the option, but use it anyways and translate it to the `estimator` equivalent. This would be somewhat backwards compatible

DavisVaughan:
Closes #163 

The yardstick global option `yardstick.event_first` is one that affects the computation of specific metric functions that compute binary classification performance. It is an example of a [hidden argument](https://design.tidyverse.org/args-hidden.html#args-hidden), and I think it should be avoided in favor of an explicit argument.

This PR introduces the explicit `event_level` argument to each class and class probability metric function that used the global option. It can be either `"first"` (the default) or `"second"`. The default is set to the internal function `yardstick_event_level()`, which will return `"first"` unless the global option of `yardstick.event_first` is set, in which case it will respect that instead since it has been manually set by the user. This is for backwards compatibility for this version of yardstick, however, a warning will be raised encouraging the user to update and in the future this support will be removed.

A few other things of note:

- The `metric_set()` function returned when class / class prob metrics are used also gains this argument

- The `summary()` method for `conf_mat` also gains this argument, solving #160.

- `mn_log_loss()` did not support the `yardstick.event_first` option. It does now.


jcpsantiago:
@DavisVaughan I'm having issues using this. Thought this was the correct way according to the help page?

```
  pr_auc_second_level <- function(data, truth, ...) {
    yardstick::pr_auc(
      data = data,
      truth = !!rlang::enquo(truth),
      ...,
      estimator = "binary",
      event_level = "second"
    )
  }
  
  pr_auc_second_level <- yardstick::new_prob_metric(pr_auc_second_level, "maximize")
  
  pr_xgb_tuned_fit <- tune::tune_grid(
    xgb_workflow,
    resamples = cv_folds,
    grid = xgb_grid,
    metrics = yardstick::metric_set(pr_auc_second_level),
    control = tune::control_grid(save_pred = TRUE, verbose = TRUE)
  )
```

PR #205, opened 2021-03-10
New metric mean error / mean signed deviation closes (#183)

### LINKED ISSUE #183: Metric: Mean Signed Deviation
The mean signed deviation is documented here : https://en.wikipedia.org/wiki/Mean_signed_deviation
This is a measure that assesses the biais of an estimator. 
This would be great to have it part of `yardstick`.

datenzauberai:
Closes #183 

This is basically the existing mpe(), but without deviding by the true value.


DavisVaughan:
@datenzauberai could you please call this mean signed deviation, and use `msd()` as the name?

datenzauberai:
Sure, though I think it's not the most obvious choice:

- 311.000 results on Google Scholar for "mean error", 501 for "mean signed deviation"
- 153 results on Cross Validated for "mean error", 1 for "mean signed deviation"
- people in the forecasting realm like Kourentzes or Hyndman use "mean error"
- "S" is normally used for squared error, see e.g. overview of error measures here http://www.gsrc.ca/metrics_typology2019.pdf

In case you would like to avoid a two letter name, I would rather suggest mbe (mean bias error) which IMO is more popular than mean signed deviation.

topepo:
I'd rather not use `me()`. Too short an ambiguous (as an acronym). Let's use `msd()` and make notes about the other names in the man file.

DavisVaughan:
@datenzauberai I'll finish this off. Thanks for the PR!

PR #222, opened 2021-05-04
POC Survival metrics for naive ROC curve and naive ROC AUC

DavisVaughan:
Still needs:

- Many tests
- Documentation
- Decision on what `_vec()` versions should do. Should the ROC AUC vec version take a `time` argument and a vector of `estimate`s (rather than the nested tibble structure)? That way it can return a single number?
- Export the ROC AUC vec variant


DavisVaughan:
It seems like survival requires R >= 3.5

mattwarkentin:
Performance metrics for survival models are near and dear to my heart/work. I have spent years of my PhD working on/with calibration and discrimination metrics for survival, competing risks, and multi-state settings.

I am wondering where would be the best place (if one exists) to engage in any discussions/planning you all might be having when thinking about how to implement such metrics for these contexts. I would love to contribute (thoughts or code).

topepo:
For performance measures, yardstick is the place. Otherwise, probably the censored package repo is best.

EmilHvitfeldt:
We ended up not going with this approach.

Issue #240, opened 2021-11-19
Access underlying data of `autoplot()` objects


joeycouse:
The `autoplot()` functions are great out of the box for quick insights, but I wish there was a way to access the underlying data used to generate said plots. For example the `autoplot.conf_mat()` does the data prep and the plotting. If users could access the underlying data we could generate custom visualizations.

The data prep isn't trivial either especially for the "mosaic" type plots. 

https://github.com/tidymodels/yardstick/blob/5f1068ff8bdd402416996da35352526a80c4c966/R/conf_mat.R#L343-L379

It would be great if there was an optional arg to just return the data or even a separate user facing function. Hadley describes this as a best practice [here](https://r-pkgs.org/r.html#isolate-side-effects).

> 7.5.3 Isolate side effects
Creating plots and printing output to the console are two other ways of affecting the global R environment. Often you can’t avoid these (because they’re important!) but it’s good practice to isolate them in functions that only produce output. This also makes it easier for other people to repurpose your work for new uses. ***For example, if you separate data preparation and plotting into two functions, others can use your data prep work (which is often the hardest part!) to create new visualisations***.

Let me know what your opinions are. I know there's a bunch of different `autoplot()` functions so any change probably warrants further discussion. Thank you!

juliasilge:
I know I would love this, because I will end up [overloading some of these methods](https://github.com/juliasilge/juliasilge.com/blob/master/content/blog/datasaurus-multiclass/index.Rmarkdown) when I want to use them in blog posts, etc.

opus1993:
I would use this feature as well to apply my fill color scale, along with prismatic to flip the color of the text labels on the dark tiles.

PR #248, opened 2021-12-10
Add `extract_plot_data()` and fill argument to `autoplot()` for type = "mosiac"

joeycouse:
Resolves part of #240 

Just putting this out there to get your thoughts on the interface. I've implemented an `extract_plot_data()` function for the confusion matrix class which returns a list with the relevant plot data. 

Would need to be extended for other `autoplot()` use cases e.g. roc_curve, etc.  If you think this is something worth merging I can put together the other methods. Just wanted to get y'alls take before putting more effort into this. Thanks!


juliasilge:
Is there some existing generic we should use for this, rather than making a new one? `fortify` comes to mind, although the [docs say not to use it](https://ggplot2.tidyverse.org/reference/fortify.html).

joeycouse:
The only function I'm aware of that achieves something similar is the `ggplot2::ggplot_build()` which accepts a plot object and returns a dataframe of the plot data. Although the returned dataframe isn't in a format I think would address #248 
adequately. 

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip

data("two_class_example")

two_class_example %>%
  conf_mat(truth, predicted) %>%
  autoplot() %>%
  ggplot_build() %>%
  pluck(1)
#> [[1]]
#>      fill  xmin  xmax       ymin       ymax PANEL group colour size linetype
#> 1 #4f58bd   0.0 258.0  0.0000000 -0.8798450     1     1     NA  0.5        1
#> 2  grey70   0.0 258.0 -0.8898450 -1.0100000     1     2     NA  0.5        1
#> 3  grey70 260.5 502.5  0.0000000 -0.2066116     1     2     NA  0.5        1
#> 4 #4f58bd 260.5 502.5 -0.2166116 -1.0100000     1     1     NA  0.5        1
#>   alpha
#> 1   0.9
#> 2   0.9
#> 3   0.9
#> 4   0.9
```

<sup>Created on 2021-12-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

topepo:
I don't think that it is a good idea to make a new generic. The `tidy()` method should translate the object to a  tabular data structure that can be used as the substrate for the `autoplot()` method. The `tidy()` method for the confusion matrix is maybe not the best (I think that I wrote it) and can be improved to make some of your (and Julia's) code more concise. 

That would be enough to get data for the heatmap but the mosaic plot would need some additional, non-tabular data. So, I propose: 

1) I'll update the PR to improve the `tidy()` method: 

2) @joeycouse can take their [work on `cm_mosaic_data()`](https://github.com/tidymodels/yardstick/pull/248/files#diff-a1a0cf15cc1c8554713b5d628b63876cb9b8dbb0f0975fe71e54f947a1514a68R350-R403) to make a function that we can export so facilitate custom mosaic plots. 

I don't think that we need `cm_heat_data()` nor do we need to export `get_axis_labels()`. 

@DavisVaughan and @juliasilge how does that sound?

juliasilge:
I think this sounds like a good way to go 👍

Issue #318, opened 2022-09-09
initial survival metrics


topepo:
Let's table the ROC curves for now. There are some methodological issues to work out. 

I think that we should start with concordance for event time outcomes and Brier scores. 

Concordance is pretty straight-forward and there is [already some work on this](https://github.com/tidymodels/yardstick/tree/feature/survival-concordance).

For  Brier scores, I would start by implementing the binary and multi class versions first (for now, let's call that function `brier_class()`). The survival version(s) (say `brier_time()`?) would need an argument for the time to compute the probability. 
There was some previous work here too #18 and #202

Some open questions: 

* How do we include the time setting in the resulting data? 
* If someone specifies multiple time points, how would a user of the tune package specify one of them to optimize on? 
* Some Brier scores need to use weights based on the probability of a data point being censored. There are a lot of ways to do this but a simple KM curve, based on the training set, would do. How would we facilitate that? 
* Brier scores can also be decomposed in a few ways. We should think about doing that in specialized functions (as in #202).

EmilHvitfeldt:
This issue is mostly done now. I'm closing it in favor of https://github.com/tidymodels/yardstick/issues/396, which is used for progress keeping

Issue #311, opened 2022-07-06
Feature request: custom `direction` in `new_metric()`


mikemahoney218:
## Feature

Apologies if this already exists and I've missed it.

It would be useful to allow custom `direction` options when creating new metrics. My current use-case for this behavior is implementing [Geary's C](https://en.wikipedia.org/wiki/Geary%27s_C) in [waywiser](https://github.com/mikemahoney218/waywiser), which has an optimal value of 1; while this case is easy to engineer around (a metric C* exists such that `C* = C - 1`, letting `direction = "zero"`) I think there may be other metrics which would benefit from being able to provide their own direction arguments.

DavisVaughan:
The biggest issue with a completely custom `direction` is that tune somehow has to know how to use it. In particular, `tune_bayes()` does different things depending on whether you are minimizing/maximizing your metric 

https://github.com/tidymodels/tune/blob/e4b7485f6196c005d457f713c2bc5da5ffbfb63c/R/tune_bayes.R#L216

So does `show_best()` I believe

EmilHvitfeldt:
Having thought about this for a while I'm going to close this issue. The gains we could have from this issue is outweighed by the  complexity we would need to add here and in tune (and friends).

Especially since there is an alternative which would be to adjust the metric in question to be either maximizable or minimizable, as suggested by @mikemahoney218 .

PR #320, opened 2022-09-13
proposal for dynamic metrics

topepo:
See #318

This PR is not meant to be merged but to serve as how to incorporate censored regression performance statistics that require a time parameter (with 1 or more values). 

The tidyup format is used. Please use this issue to discuss. Potentially interested parties: @mikemahoney218, @dincerti, and @exuberantleigh

### DIFF
--- a/dynamic metrics.md
+++ b/dynamic metrics.md
@@ -1,0 +1,165 @@
+# yardstick modifications for censored data metrics
+
+**Champion**: Max Kuhn  
+**Co-champion**: Hannah Frick, Emil Hvitfeldt  
+**Status**: Proposal  
+
+## Abstract
+
+Censored regression models have an added dimension to their analysis. To compute model performance, some metrics require the specification of _analysis times_ (a.k.a. time prediction horizons). Thgesr are chosen by the user and depend on the data and the context of their application. For example, if 5- and 10-year cancer-free survival are clinically important quantities, the model's performance should be evaluated on how well it predicts at the 5 and 10 year marks. 
+
+This requires some changes in tidymodels, specifically yardstick performance metrics. 
+
+## Motivation
+
+With censored time-to-event data, there are performance metrics that are _dynamic_ in nature; one or more time points for the analysis must be specified. For example, the [Brier score](https://en.wikipedia.org/wiki/Brier_score) for [censored data](https://arxiv.org/abs/1912.08581) uses survivor probability predictions at specific time points. Users will typically want dynamic metrics computed at multiple points in time. 
+
+The results of a yardstick function or metrics set currently contain three columns: `.metric` (character), `.estimator` (double), and `.estimate` (double). There is no available slot for ancillary information, such as the time-of-analysis. We will suggest how to include this information into metric results.
+
+
+## Solution
+
+(Notation: in the discussion here, the vector of time points for the dynamic analyses will be denoted as `.time`.) 
+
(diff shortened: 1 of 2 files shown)


dynamic metrics.md
```diff
@@ -0,0 +1,165 @@
...
+
+There very well might be other situations in the future where there are additional parameters (similar to `.time`). We believe that a list column approach may be used to address these issues. 
+
+Two potential uses are: 
+
+* For some spatial metrics, additional _data_ must be attached to compute additional metrics. The documentation for the [waywiser](https://github.com/mikemahoney218/waywiser) package shows an example where spatial geometry data is returned  to calculate spatial weights.
+
+* There is a tidymodels proposal to compute model characterization values to optionally return with performance metrics. For example, for a simple decision tree, we might return the number of terminal nodes, the number of predictors used in the splits, and the names of these predictors. While the first two quantities are scalars, the last is a vector of predictor names.
+
+It is possible that the approach listed above could accommodate situations such as these.
```
mikemahoney218, line 119: I don't immediately understand how this would extend to work with spatial data. In that situation, you need the actual geometry for each row, which is an (effectively) unique identifier. You also need _all_ of these values (for the assessment set, at least) at the same time; the metrics depend on the relative positions of each point in the assessment set to each other, not just the position of the current point.

That said, I've started thinking that the actual workflow in practice for spatial metrics isn't going to be "look at the mean spatial metrics for each assessment set in fit_resamples", but rather "connect the predictions from fit_resamples back to the original data and estimate the spatial metrics once, for the entire data set, using hold-out predictions". That's not going to require any changes in yardstick, I _think_, and so I think it's fine to do the optimal thing for survival analyses without worrying about spatial data.

mikemahoney218: > I've started thinking that the actual workflow in practice for spatial metrics...

For context: no one is really tuning models based on spatial error distributions (for the same reason you don't tune based on bias, as it doesn't improve accuracy); they're checking spatial autocorrelation to either justify ignoring spatial structure in their model or to identify "problem areas" and potential ways to mitigate them. I think a more common workflow is where parameters are tuned using `rmse()`, and then spatial autocorrelation is checked before the model is "finalized".

dynamic metrics.md
```diff
@@ -0,0 +1,165 @@
...
+Two potential uses are: 
+
+* For some spatial metrics, additional _data_ must be attached to compute additional metrics. The documentation for the [waywiser](https://github.com/mikemahoney218/waywiser) package shows an example where spatial geometry data is returned  to calculate spatial weights.
+
+* There is a tidymodels proposal to compute model characterization values to optionally return with performance metrics. For example, for a simple decision tree, we might return the number of terminal nodes, the number of predictors used in the splits, and the names of these predictors. While the first two quantities are scalars, the last is a vector of predictor names.
+
+It is possible that the approach listed above could accommodate situations such as these. 
+
+
+## Other issues issues around dynamic metrics
```
mikemahoney218, line 122: I really like this section -- passing multiple weight-matrix functions through `.weights` feels great, and I'd love for passing `select_best_*` to be supported to solve #311

DavisVaughan (approved):
This aligns with my interpretation of what we discussed in the group meeting, and still feels right to me

(Approving to show I agree with what is written, not to request that it is merged)

EmilHvitfeldt (approved):

hfrick (approved):

simonpcouch (approved):

EmilHvitfeldt:
This PR has served its purpose, I'm closing it.

PR #321, opened 2022-09-27
update snapshot of internal yardstick_table

EmilHvitfeldt:
The new {rlang} version introduced links in internal errors https://github.com/r-lib/rlang/commit/a72c03c5c9e9d72c94c27082d5298a394577af69. This PR updates the 5 affected snapshots

### DIFF
--- a/tests/testthat/_snaps/misc.md
+++ b/tests/testthat/_snaps/misc.md
@@ -5,7 +5,8 @@
     Condition
       Error in `yardstick_table()`:
       ! `truth` must be a factor.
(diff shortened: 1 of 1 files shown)


DavisVaughan:
@EmilHvitfeldt when I update snapshots to take advantage of a new feature in a dependency, I always require that version of the dependency as the minimum version in Imports, i.e. `rlang (>= 1.0.6)`. 

I find that this has 0 negative impact on users, and makes it easier for any developers that work on the package because devtools will ensure that everyone has the correct package versions to run the tests cleanly

PR #322, opened 2022-09-27
More metric summarizer

EmilHvitfeldt:
This PR adds a refactoring of `metric_summarizer()` by splitting it into 3 new functions `metric_summarizer_class()`, `metric_summarizer_numeric()`, and `metric_summarizer_prob()`. (`metric_summarizer()` is being soft deprecated)

This will lessen the burden of modifying any one of the types of metrics as `metric_summarizer()` was forced to cover all cases. This PR is one step towards adding survival metrics.

`metric_summarizer_class()` and `metric_summarizer_prob()` are carbon copies of `metric_summarizer()`, where on the other hand , `metric_summarizer_numeric()` doesn't have an `estimator` or `event_level` arguments as they aren't used for numeric metrics.

I decided to document them separately, as I found it less confusing to have `for numeric metrics this is NULL, for class metrics ...`.

PR #328, opened 2022-10-17
Draft POC for tidyselect summarizers

DavisVaughan:


DavisVaughan:
@EmilHvitfeldt okay this is ready for you to take a look at and pull pieces into #322 

It should actually be in a state where it passes CRAN checks, so it will mostly be a job of carefully copy/pasting into your current PR on top of the work you have done

If you can work out the merge conflicts, you could also rebase this PR onto that one and merge this one straight in, but that might be a little tricky.

Probably also needs a NEWS bullet, something like:

```r
* yardstick metric functions now use a pure tidyselect interface for `truth`, `estimate`, and the `...` of class probability metrics (#322).
```

tests/testthat/_snaps/template.md
```diff
@@ -84,35 +80,30 @@
...
+        truth = obs, VF:L, obviouslywrong = TRUE)
     Condition
-      Error in `prob_metric_summarizer()`:
-      ! `...` must be empty.
-      x Problematic argument:
-      * obviouslywrong = TRUE
+      Error:
+      ! Must subset columns with a valid subscript vector.
+      x Subscript `TRUE` has the wrong type `logical`.
+      i It must be numeric or character.
```
DavisVaughan, line 108: Weird case where `obviouslywrong = TRUE` actually gets passed through the `...` to the tidyselect function, but I'm not worried about this

tests/testthat/test-template.R
```diff
@@ -474,7 +475,7 @@ test_that("prob_metric_summarizer()'s na_rm argument work", {
     fn = roc_auc_vec,
     data = hpc_f1_na,
     truth = obs,
-    estimate = matrix(data = c(VF = VF, F = F, M = M, L = L), ncol = 4L),
+    VF:L,
```
DavisVaughan, line 478: New interface makes for easier testing for sure

EmilHvitfeldt:
Thanks for all the work! close with https://github.com/tidymodels/yardstick/pull/322

PR #337, opened 2022-10-26
New metric vec template

EmilHvitfeldt:
This PR adds a refactoring of `metric_vec_template()` into `numeric_metric_vec_template()`, `class_metric_vec_template()` and `prob_metric_vec_template()`. 

At the same time it also refactors `validate_truth_estimate_checks()` into `validate_factor_truth_factor_estimate()`, `validate_factor_truth_metrix_estimate()`, `validate_numeric_truth_numeric_estimate()`, and `validate_binary_estimator()` which are being called inside `*_vec()` functions.

Old functions are soft deprecated


EmilHvitfeldt:
@DavisVaughan You are NOT allowed to look at this yet! I need to sleep and do another pass

R/template.R
```diff
@@ -266,133 +265,161 @@ prob_estimate_convert <- function(estimate) {
...
 #' @export
-metric_vec_template <- function(metric_impl,
-                                truth,
-                                estimate,
-                                na_rm = TRUE,
-                                cls = "numeric",
-                                estimator = NULL,
-                                case_weights = NULL,
-                                ...) {
+numeric_metric_vec_template <- function(metric_impl,
```
DavisVaughan, line 318: It is interesting to me that the vec template functions now have 4 purposes:

- Validate case weights
- Convert from class_pred->factor as needed
- Handle missing values. Importantly, return `NA_real_` as needed rather than calling the `metric_impl`.
- Call the `metric_impl()`

The first 2 could easily be wrapped up in helpers that we'd expect the metric functions to call themselves (you already have `validate_case_weights()` and the other could be `truth <- convert_class_pred(truth)`)

The 3rd could be put into its own helper too, basically `handle_na_rm()` that I already mentioned, but this is a little trickier because the return value is somewhat complicated and the metric functions would have to correctly return `NA_real_` themselves if `handle_na_rm()$return_missing` was true

---

There are a few really big upsides of this though:
- You wouldn't need these template functions _at all_. You'd put slightly more burden on the metric authors to call the `validate_*()` and `handle_na_rm()` functions manually, but you'd remove a ton of pain related to having to call the `metric_impl()` function for them.
- Metric function authors wouldn't need to do the weird function-within-function thing. Like defining `j_index_impl()` inside `j_index_vec()`. If they are in charge of calling `j_index_impl()`, it can be put anywhere and could have `estimator, event_level` as arguments that could be manually passed through. That would be way simpler and easier to explain.
- It would also be way more flexible because metric authors wouldn't be boxed in by our validation checking code. If they don't want it or need something else, they can just roll their own

It might be worth trying the idea out on a few metrics to see if it would be worth it. I think the toughest part is just handling `handle_na_rm()`, but I think the pattern used in every metric would be

```r
res <- handle_na_rm(truth, estimate, case_weights, na_rm)
truth <- res$truth
estimate <- res$estimate
case_weights <- res$case_weights

if (res$return_missing) {
  return NA_real_
}

j_index_impl(truth, estimate, case_weights, estimator, event_level)
```

which isn't really that bad

R/estimator-helpers.R
```diff
@@ -87,7 +87,7 @@ get_weights <- function(data, estimator) {
 #' @param estimator Either `NULL` for auto-selection, or a single character
 #' for the type of estimator to use.
 #'
-#' @seealso [class_metric_summarizer()] [metric_vec_template()]
+#' @seealso [metric-summarizers()] [metric-vec_template()]
```
DavisVaughan, line 90: Do these links really work with the `-` and the `()`?

R/prob-helpers.R
```diff
@@ -44,8 +44,8 @@ auc <- function(x, y, na_rm = TRUE) {
 #' name if 1 input is provided to `...` or it constructs a quosure where the
 #' expression constructs a matrix of as many columns as are provided to `...`.
 #' These are eventually evaluated in the `summarise()` call in
-#' [metric_summarizer()] and evaluate to either a vector or a matrix for further
-#' use in the underlying vector functions.
+#' [metric-summarizers()] and evaluate to either a vector or a matrix for
```
DavisVaughan, line 47: Same question here about `-` and `()`. Might want to Find in Files for that

R/template.R
```diff
@@ -266,133 +265,161 @@ prob_estimate_convert <- function(estimate) {
...
-#' type should be finalized, so this should be a character vector of length 1\.
-#' By default, this character value is required to be one of: `"binary"`,
-#' `"macro"`, `"micro"`, or `"macro_weighted"`. If your metric allows more
-#' or less averaging methods, override this with `averaging_override`.
-#'
-#' @param case_weights Optionally, the realized case weights, as a numeric
-#' vector. This must be the same length as `truth`, and will be considered in
-#' the `na_rm` checks. If supplied, this will be passed on to `metric_impl` as
-#' the named argument `case_weights`.
+#' @param case_weights the realized case weights, as a numeric vector. This must
```
DavisVaughan, line 293:
```suggestion
#' @param case_weights The realized case weights, as a numeric vector. This must
```

R/template.R
```diff
@@ -266,133 +265,161 @@ prob_estimate_convert <- function(estimate) {
...
+#'   be the same length as `truth`, and will be considered in the `na_rm`
+#'   checks. If supplied, this will be passed on to `metric_impl` as the named
+#'   argument `case_weights`.
```
DavisVaughan, line 296: Not optional anymore right?

R/template.R
```diff
@@ -266,133 +265,161 @@ prob_estimate_convert <- function(estimate) {
...
 #'
 #' @param ... Extra arguments to your core metric function, `metric_impl`, can
-#' technically be passed here, but generally the extra args are added through
-#' R's scoping rules because the core metric function is created on the fly
-#' when the vector method is called.
+#'   technically be passed here, but generally the extra args are added through
+#'   R's scoping rules because the core metric function is created on the fly
+#'   when the vector method is called.
```
DavisVaughan, line 301: This is the craziness you'd be able to eliminate by letting the metric functions call their "core metric functions" themselves. Wouldn't need any reliance on scoping rules, and would be way more R like

R/template.R
```diff
@@ -266,133 +265,161 @@ prob_estimate_convert <- function(estimate) {
...
-metric_vec_template <- function(metric_impl,
-                                truth,
-                                estimate,
-                                na_rm = TRUE,
-                                cls = "numeric",
-                                estimator = NULL,
-                                case_weights = NULL,
-                                ...) {
+numeric_metric_vec_template <- function(metric_impl,
+                                        truth,
```
DavisVaughan, line 319: I'd also consider naming any new exported validation functions with a `check_*()` prefix. That is what we use in rlang, vctrs, dplyr, tidyr, etc, and is becoming the standard prefix for things like this

R/validation.R
```diff
@@ -173,35 +83,72 @@ validate_truth_estimate_lengths <- function(truth, estimate) {
...
   }
-}
 
-validate_truth_estimate_checks <- function(truth, estimate,
-                                           cls = "numeric",
-                                           estimator) {
+  if (estimator == "binary") {
+    if (is.matrix(estimate)) {
+      abort(paste0(
+        "You are using a `binary` metric but have passed multiple columns to `...`"
```
DavisVaughan, line 101:
```suggestion
        "You are using a binary metric but have passed multiple columns to `...`."
```

R/validation.R
```diff
@@ -253,3 +200,24 @@ validate_estimator <- function(estimator, estimator_override = NULL) {
...
 }
+
+validate_case_weights <- function(case_weights, size) {
+  if (is.null(case_weights)) {
+    return(invisible())
+  }
+
+  if (!is.integer(case_weights) && !is.double(case_weights)) {
+    abort("`case_weights` must be an integer or double vector.")
+  }
```
DavisVaughan, line 211: Was I doing any validation of the `case_weights` type before hand? I think this technically works if people provide hardhat's case weight types, but ideally `is.integer()` would actually return `FALSE` for those. It happens to return `TRUE` because those are the storage types of the vectors.

I think I was previously just letting `yardstick_table()` -> `hardhat::weighted_table()` handle the validation of the `case_weights` because it uses `vec_cast()`

DavisVaughan (commented):
I think this is definitely moving in the right direction. I think this idea of moving validation into the metric functions themselves is really smart, and very powerful!

I provided some fairly high level feedback. My main thought (which I have put in detail below) is that I'm wondering if we can make this _even_ simpler, more flexible, and more robust by breaking down the jobs performed by the 3 new vec-template functions into multiple helper functions that we'd force the metric functions to call themselves.

I think I was a bit obsessed with trying to shoehorn everything into 1 `metric_vec_template()` function, but that has proven to be pretty strict and has kept some people from being able to write their extension functions.

Instead, exporting a number of optional check/validate/na_rm functions that authors could use in their own metrics as needed feels much less restrictive.

As I mention below, it also has the massive benefit of not needing to write your `*_impl()` function _inside_ the `*_vec()` function (because we needed to use R's scoping rules weirdly), which I think makes this heavily worth considering

We can schedule some time to talk ASAP if you want to discuss this further

R/estimator-helpers.R
EmilHvitfeldt: They match the `@name` for the summarizer functions. I opted to use the `-` instead of `_` because it becomes very close to the deprecated `metric_summarizer()`

Kapture 2022-11-22 at 15 49 55

NAMESPACE
```diff
@@ -92,6 +92,7 @@ export(bal_accuracy_vec)
 export(ccc)
```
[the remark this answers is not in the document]
EmilHvitfeldt: It should be. Now that we don't force any metric to be validated, the developer can write the validation that is correct for that metric.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # yardstick (development version)
 
+* `metric_vec_template()` is being soft deprecated in favor of the more specific
+  newly added `class_metric_vec_template()`, `numeric_metric_vec_template()`, 
+  and `prob_metric_vec_template()`.
+
```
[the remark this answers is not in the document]
EmilHvitfeldt: exported

EmilHvitfeldt:
I did a small rewrite since our design

```r
if (na_rm) {
    result <- handle_missings(truth, estimate, case_weights)

    if (result$na_exit) {
      return(NA_real_)
    }

    truth <- result$truth
    estimate <- result$estimate
    case_weights <- result$case_weights
  }
```
didn't work since there is no way to hit the return statement in a meaningful way. It now reads

```r
if (na_rm) {
    result <- handle_missings(truth, estimate, case_weights)

    truth <- result$truth
    estimate <- result$estimate
    case_weights <- result$case_weights
  } else {
    any_na <- detect_missings(truth, estimate, case_weights)
    if (any_na) {
      return(NA_real_)
    }
  }
```

R/validation.R
```diff
@@ -173,49 +102,78 @@ validate_truth_estimate_lengths <- function(truth, estimate) {
...
+      ))
+    }
+  } else {
+    n_lvls <- length(levels(truth))
+    n_cols <- ncol(estimate)
+
+    if (is.null(n_cols) || n_lvls != n_cols) {
+      if (is.null(n_cols)) {
+        n_cols <- 1
+      }
```
DavisVaughan, line 147:
```suggestion
    if (is.matrix(estimate)) {
      n_cols <- ncol(estimate)
    } else {
      n_cols <- 1L
    }
    

    if (n_lvls != n_cols) {
```

Maybe simpler?

DavisVaughan (approved):
A number of comments about naming, and a request to open a few issues related to supporting class-pred in a better way and some issues with the curve functions, but overall seems like a really nice simplification, so I'm going to approve alongside this feedback

After addressing my comments, I would encourage you to do one big self review for grammar and documentation consistency, as this touches a lot of files and it is easy to get things out of sync

NAMESPACE
```diff
@@ -106,6 +110,7 @@ export(gain_capture)
 export(gain_capture_vec)
 export(gain_curve)
 export(get_weights)
+export(handle_missings)
```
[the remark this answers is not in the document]
DavisVaughan: See my other comment about adding a `yardstick_` prefix

NAMESPACE
```diff
@@ -91,10 +91,14 @@ export(bal_accuracy)
...
 export(ccc)
 export(ccc_vec)
+export(check_class_metric)
+export(check_numeric_metric)
+export(check_prob_metric)
 export(class_metric_summarizer)
 export(classification_cost)
 export(classification_cost_vec)
 export(conf_mat)
+export(detect_missings)
```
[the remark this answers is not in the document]
DavisVaughan: See my other comment about adding a `yardstick_` prefix

NEWS.md
```diff
@@ -1,5 +1,19 @@
...
+  and flexible metric creation approach. `handle_missings()` and 
+  `detect_missings()` have been added for treatment of missing values. 
+  `check_class_metric()`, `check_numeric_metric()`,
+  and `check_prob_metric()` have been added to perform standardized input
+  checking classification, regression and class probability metrics.   
+  `validate_factor_truth_factor_estimate()`,
+  `validate_factor_truth_metrix_estimate()`,
+  `validate_numeric_truth_numeric_estimate()`, `validate_binary_estimator()` and
+  `validate_case_weights()` have been also been exposed for more granular 
+  validation then the `check_*_metric()` functions. These changes mean that it
```
DavisVaughan, line 13: Update this after un-exporting these

NEWS.md
```diff
@@ -1,5 +1,19 @@
...
+  `detect_missings()` have been added for treatment of missing values. 
+  `check_class_metric()`, `check_numeric_metric()`,
+  and `check_prob_metric()` have been added to perform standardized input
+  checking classification, regression and class probability metrics.   
+  `validate_factor_truth_factor_estimate()`,
+  `validate_factor_truth_metrix_estimate()`,
+  `validate_numeric_truth_numeric_estimate()`, `validate_binary_estimator()` and
+  `validate_case_weights()` have been also been exposed for more granular 
+  validation then the `check_*_metric()` functions. These changes mean that it 
+  is the developer's responsibility to perform validation of `turth` and
```
DavisVaughan, line 14:
```suggestion
  is the developer's responsibility to perform validation of `truth` and 
```

NAMESPACE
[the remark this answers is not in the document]
EmilHvitfeldt: https://github.com/tidymodels/yardstick/pull/337/commits/03c13555a620e6158eec672ada89bfb0217e2f64

NAMESPACE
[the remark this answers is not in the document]
EmilHvitfeldt: https://github.com/tidymodels/yardstick/pull/337/commits/cc828152f041bc9b2990b985bc41c359d88516d7

NAMESPACE
```diff
@@ -182,7 +187,12 @@ export(spec_vec)
 export(specificity)
 export(specificity_vec)
 export(tidy)
+export(validate_binary_estimator)
+export(validate_case_weights)
 export(validate_estimator)
+export(validate_factor_truth_factor_estimate)
+export(validate_factor_truth_matrix_estimate)
+export(validate_numeric_truth_numeric_estimate)
```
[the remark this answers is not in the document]
EmilHvitfeldt: https://github.com/tidymodels/yardstick/pull/337/commits/76be37bba980d524e507c4e9a1d14ec42493cec0

R/check_metric.R
```diff
@@ -0,0 +1,51 @@
+#' Developer function for checking inputs in new metrics
```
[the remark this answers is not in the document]
EmilHvitfeldt: https://github.com/tidymodels/yardstick/pull/337/commits/bc01a40d8d5c5d63c3f69045c40f2803d941c9d9

R/check_metric.R
```diff
@@ -0,0 +1,51 @@
...
+#'
+#' `check_numeric_metric()`, `check_class_metric()`, and `check_prob_metric()`
+#' are useful alongside the [metric-summarizers()] functions for implementing
```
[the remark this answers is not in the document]
EmilHvitfeldt: https://github.com/tidymodels/yardstick/pull/337/commits/30df9b4a6cc148b528fc30e4334bb8720975ce83

R/metrics.R
```diff
@@ -82,6 +82,13 @@ metrics.data.frame <- function(data,
   estimate <- tidyselect::vars_pull(names, {{estimate}})
   probs <- unname(tidyselect::vars_select(names, ...))
 
+  if (is_class_pred(data[[truth]])) {
+    data[[truth]] <- as_factor_from_class_pred(data[[truth]])
+  }
+  if (is_class_pred(data[[estimate]])) {
+    data[[estimate]] <- as_factor_from_class_pred(data[[estimate]])
+  }
```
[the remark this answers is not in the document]
EmilHvitfeldt: all the class_pred code in this PR is to keep tests passing

PR #340, opened 2022-11-14
consistent aspect ratio for PR curves

topepo:
All of our probability curve `autoplot()` methods use an aspect ratio of 1.00 except PR curves. They all have aces that are on 0:1 and it makes sense to keep them squares.


EmilHvitfeldt:
looks good!

PR #356, opened 2023-02-06
Curve metric summarizer

EmilHvitfeldt:
With the changes to dplyr, `summarize()` now produces a warning for multi row output. We got this in all the curve metrics.

This PR adds `curve_metric_summarizer()` which works almost identical to `prob_metirc_summarizer()` except that it uses `reframe()` instead of `summarize()`.


EmilHvitfeldt:
This PR is expected to fail CI as it doesn't fix some dplyr warnings that changes the snapshots

Issue #327, opened 2022-10-12
Error in `yardstick_table()`: ! `truth` and `estimate` must have the same levels in the same order.


deschen1:
I'm doing as I'm told by the yardstick function and posting this issue here.

When using the iris data set, it seems yardstick gets confused because in my code I turn the predicted class into a factor of 1, 2, 3 rather tahn the original character factor (with levels "setosa"...).

So the source of error is clear, but I'm posting it here nonetheless because the function asked me to. So maybe you either want to allow `conf_mat` to work with non-identical factor levels or maybe need to make this "posting request" a bit more tight so that it doesn't pop up for such an issue.

```r
iris

model_recipe <- recipes::recipe(Species ~ ., data = iris)

# Create a workflow
model_final <- parsnip::naive_Bayes(Laplace = 1) |>
  parsnip::set_mode("classification") |>
  parsnip::set_engine("klaR",
                      prior = rep(1/3, 3),
                      usekernel = FALSE)

model_final_wf <- workflows::workflow() |>
  workflows::add_recipe(model_recipe) |>
  workflows::add_model(model_final)

train_fit <- model_final_wf |>
  generics::fit(data = iris)

# Add predictions, in the mutate for class_pred the new [1, 2, 3] factor is introduced
train_predictions <- predict(train_fit, iris, type = "prob") |> 
  dplyr::mutate(class_pred = as.factor(apply(dplyr::across(tidyselect::everything()), 1, which.max))) |> 
  dplyr::bind_cols(iris)

# See the problem
head(train_predictions)
# A tibble: 6 × 9
  .pred_setosa .pred_versicolor .pred_virginica class_pred Sepal.Length Sepal.Width Petal.Length Petal.Width Species
         <dbl>            <dbl>           <dbl> <fct>             <dbl>       <dbl>        <dbl>       <dbl> <fct>  
1         1            2.98e-18        2.15e-25 1                   5.1         3.5          1.4         0.2 setosa 
2         1            3.17e-17        6.94e-25 1                   4.9         3            1.4         0.2 setosa 
3         1            2.37e-18        7.24e-26 1                   4.7         3.2          1.3         0.2 setosa 
4         1            3.07e-17        8.69e-25 1                   4.6         3.1          1.5         0.2 setosa 
5         1            1.02e-18        8.89e-26 1                   5           3.6          1.4         0.2 setosa 
6         1.00         2.72e-14        4.34e-21 1                   5.4         3.9          1.7         0.4 setosa 

[...]

Issue #314, opened 2022-08-26
Internal error in `conf_mat()`


DavisVaughan:
We just need to check that both truth/estimate are factors. Also maybe check the type of case_weights if they are present

``` r
yardstick::conf_mat(mtcars, mpg, cyl)
#> Error in `yardstick_table()`:
#> ! `truth` must be a factor.
#> ℹ This is an internal error in the yardstick package, please report it to the package authors.

yardstick::sens(mtcars, mpg, cyl)
#> Error in `dplyr::summarise()`:
#> ! Problem while computing `.estimate = metric_fn(truth = mpg, estimate =
#>   cyl, na_rm = na_rm, event_level = "first")`.
#> Caused by error in `validate_class()`:
#> ! `truth` should be a factor but a numeric was supplied.
```

<sup>Created on 2022-08-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Issue #234, opened 2021-09-02
Same syntax, same arguments to conf_mat() and roc_auc() fails with roc_auc(), and inconsistent behavior in roc_auc() arguments


patc3-phd:
## The problem

The function `conf_mat()` and `roc_auc()` have the same arguments (data, truth, estimate), and I expected them to behave similarly. However I was using character variables to refer to column names (like `target` and `prediction` to refer to the target and prediction variables). 

With `conf_mat()`, you can pass these variables as is, or use `get(variable)` or `!!variable`.

With `roc_auc()` however, the behavior is inconsistent: you can pass the variable to the `truth` argument using either `!!variable` or `get(variable)`, but not simply `variable`, in which case you get the error `truth should be a factor but a character was supplied.` Meanwhile, you can pass the variable to the `estimate` argument using either `!!variable` or as is `variable`, but not `get(variable)`, in which case you get the error `object 'prob' not found`.

The error message that accompanies `truth = variable` in particular gave me a headache, as it says truth should be a factor, and that's what I had.

So I see two oddities:

1. Inconsistent behavior between `conf_mat()` and `roc_auc()` when using variables to refer to variable names, even though they have the same format
2. Inconsistent behavior between the arguments of `roc_auc()` when using variables to refer to variable names

## Reproducible example

```r
library(tidymodels)
#> Warning: package 'tidymodels' was built under R version 4.1.1
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip
#> Warning: package 'dials' was built under R version 4.1.1
#> Warning: package 'infer' was built under R version 4.1.1
#> Warning: package 'modeldata' was built under R version 4.1.1
#> Warning: package 'parsnip' was built under R version 4.1.1
#> Warning: package 'recipes' was built under R version 4.1.1
#> Warning: package 'rsample' was built under R version 4.1.1
#> Warning: package 'tune' was built under R version 4.1.1
#> Warning: package 'workflows' was built under R version 4.1.1
#> Warning: package 'workflowsets' was built under R version 4.1.1
#> Warning: package 'yardstick' was built under R version 4.1.1

# df
df<-data.frame(
  y=factor(c("A", "A", "B", "B", "B", "B")),
  pred=factor(c("A", "B", "A", "B", "B", "B")),
  prob=c(.88,.65,.7,.22,.44,.13)
)

# variables to refer to column names
target<-"y"
prediction<-"pred"
proba<-"prob"
[...]

PR #362, opened 2023-02-07
Add conf_mat_impl()

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/yardstick/issues/327, close https://github.com/tidymodels/yardstick/issues/314, and close https://github.com/tidymodels/yardstick/issues/234.

Users were getting internal errors when using `conf_mat()` incorrectly. This happened because the `conf_mat()` functions called `yardstick_table()` directly without doing any input checking.

This PR fixes that by adding the function `conf_mat_impl()` that is used internally in `conf_mat.data.frame()` and `conf_mat.grouped_df()` such that we can do input checking correctly.

### DIFF
--- a/R/conf_mat.R
+++ b/R/conf_mat.R
@@ -176,6 +176,21 @@ conf_mat.grouped_df <- function(data,
   )
 }
 
+conf_mat_impl <- function(truth, estimate, case_weights) {
+  estimator <- "not binary"
+  check_class_metric(truth, estimate, case_weights, estimator)
+
+  if (length(levels(truth)) < 2) {
+    abort("`truth` must have at least 2 factor levels.")
+  }
+
+  yardstick_table(
+    truth = truth,
+    estimate = estimate,
+    case_weights = case_weights
+  )
+}
+
 #' @export
 conf_mat.table <- function(data, ...) {
   check_table(data)
--- a/R/conf_mat.R
+++ b/R/conf_mat.R
@@ -102,19 +102,36 @@ conf_mat.data.frame <- function(data,
     warn_conf_mat_dots_deprecated()
   }
 
-  names <- names(data)
+  truth <- enquo(truth)
+  estimate <- enquo(estimate)
+  case_weights <- enquo(case_weights)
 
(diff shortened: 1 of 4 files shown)


R/conf_mat.R
```diff
@@ -102,23 +102,40 @@ conf_mat.data.frame <- function(data,
...
+  truth <- enquo(truth)
+  estimate <- enquo(estimate)
+  case_weights <- enquo(case_weights)
 
-  truth <- tidyselect::vars_pull(names, {{truth}})
+  truth <- yardstick_eval_select(
+    expr = truth,
+    data = data,
+    arg = "truth",
+    error_call = error_call
```
DavisVaughan, line 113: I don't think you have something called `error_call` here?

DavisVaughan: You can probably just remove this line and let the default arg do its job

R/conf_mat.R
```diff
@@ -142,28 +159,40 @@ conf_mat.grouped_df <- function(data,
...
   dplyr::summarise(
     data,
     conf_mat = {
-      table <- yardstick_table(
-        truth = !!truth,
-        estimate = !!estimate,
+      table <- conf_mat_impl(
+        truth = .data[[truth]],
+        estimate = .data[[estimate]],
         case_weights = !!case_weights
```
DavisVaughan, line 196:
```suggestion
        case_weights = .data[[!!case_weights]]
```

Would this have worked? Instead of using `expr()` above?

DavisVaughan: Maybe you don't even need the `!!`?

DavisVaughan (approved):
Seems right to me

R/conf_mat.R
DavisVaughan: If you add a test where tidyselection fails (like if the column doesnt exist in the data) then you'd catch the bug here

R/conf_mat.R
EmilHvitfeldt: Couldn't find a good way to make it happen

PR #347, opened 2022-11-29
WIP Dynamic metrics for survival metrics

EmilHvitfeldt:
This PR is still work in progress. It establishes the core internal functions needed to implement survival metrics

What needs to happen:
- [x] Add `*_metric_summarizer()`
    - [x] dynamic
        - [x] Figure out how to make `yardstick_remove_missing()` work with list columns
        - [x] Figure out how to make `yardstick_remove_missing()` work with Surv objects...
    - [x] static
- [x] Add `check_*()`
    - [x] dynamic
    - [x] static
- [x] Add `brier_survival()` (dynamic)
    - [x] infrastructure
    - [x] actual correct calculations
- [x] Add `surv_concordance()` (static)
    - [x] infrastructure
    - [x] actual correct calculations
    - [x] dont use `survival::concordance()`
- [x] add tests
    - [x] `dynamic_survival_metric_summarizer()` 
    - [x] `static_survival_metric_summarizer()`
- [x] figure out how to handle `.time` that doesn't have calculated `.pred_survival`

Write function that `surv object, .time, prob weighting → factor`

- [ ] don't use `survival::concordance()` in `concordance_survival()`
- [ ] make case_weights work correctly in `brier_survival()` (right now it doesn't normalize correctly

``` r
library(yardstick)

data("lung_surv")

lung_surv %>%
  concordance_survival(surv_obj, age)
#> # A tibble: 1 × 3
#>   .metric              .estimator .estimate
#>   <chr>                <chr>          <dbl>
#> 1 concordance_survival binary         0.484

lung_surv %>%
  brier_survival(surv_obj, .pred, prob_censored, .time = c(100, 500, 1000)) %>%
  pull(.estimate)
#> [[1]]
#> # A tibble: 3 × 2
#>   .time .estimate
#>   <dbl>     <dbl>
#> 1   100    0.163 
#> 2   500    0.237 
#> 3  1000    0.0676

# Metric sets work
metric_set(brier_survival, brier_survival)(
  lung_surv, surv_obj, .pred, prob_censored, .time = c(100, 500, 1000)
)
#> # A tibble: 2 × 3
#>   .metric        .estimator .estimate       
#>   <chr>          <chr>      <list>          
#> 1 brier_survival binary     <tibble [3 × 2]>
#> 2 brier_survival binary     <tibble [3 × 2]>
```

PR #374, opened 2023-02-27
Brier survival

EmilHvitfeldt:
This PR is the third in a series of PRs were I chip away at https://github.com/tidymodels/yardstick/pull/347, for ease of reviewing.

This PR:

- Adds `brier_survival()`
- Sets up result comparison with python {sksurv}


R/surv-brier.R
```diff
@@ -0,0 +1,166 @@
...
+    norm_const <- sum(case_weights)
+  } else {
+    case_weights <- rep(1, length(estimate))
+    norm_const <- sum(!survival::is.na.Surv(truth))
+  }
+
+  category_1 <- surv_time < eval_time & surv_status == 1
+  category_2 <- surv_time >= eval_time
+
+  # (0 - estimate) ^ 2 == estimate ^ 2
```
hfrick, line 159: 😄

R/surv-brier.R
```diff
@@ -0,0 +1,166 @@
...
+#' [quasiquotation][rlang::quasiquotation] (you can unquote column names). For
+#' `_vec()` functions, a numeric vector.
+#'
+#' @param eval_time The column identifier for the time point. This
+#' should be a numeric vector, with 1 unique value for each group. This should
+#' be an unquoted column name although this argument is passed by expression and
+#' supports [quasiquotation][rlang::quasiquotation] (you can unquote column
+#' names). For `_vec()` functions, a numeric vector.
+#'
+#' @param ... Not currently used.
```
hfrick, line 44: we could back that up with a `rlang::check_dots_empty()`

man-roxygen/return-dynamic-survival.R
```diff
@@ -0,0 +1,10 @@
+#' @return
+#'
+#' A `tibble` with columns `.metric`, `.estimator`,
+#' and `.estimate` and 1 row of values.
+#'
+#' For grouped data frames, the number of rows returned will be the same as
+#' the number of groups.
```
hfrick, line 7:
```suggestion
#' A `tibble` with columns `.metric`, `.estimator`, and `.estimate`.
#'
#' For an ungrouped data frame, the result has one row of values. For a grouped data frame, 
#' the number of rows returned is the same as the number of groups.
```

tests/testthat/test-surv-brier.R
```diff
@@ -0,0 +1,49 @@
...
+    brier_res_case_wts$.estimate
+  )
+})
+
+# sklearn compare --------------------------------------------------------------
+
+test_that('sklearn equivalent', {
+  py_res <- read_pydata("py-brier-survival")
+
+  eval_times <- c(100, 500, 1000)
```
hfrick, line 33: I think that's a leftover line, you don't seem to need it (anymore).

tests/testthat/test-surv-brier.R
```diff
@@ -0,0 +1,49 @@
...
+
+  lung_surv <- data_lung_surv()
```
hfrick, line 35: An out-of-scope comment:
I don't quite understand what the purpose of that function is. Is that so that the dataset only exists within the environment of that call to `test_that()`? 

If so, you could use the `envir` argument to `data` à la
```
data(time_to_million, package = "censored", envir = rlang::current_env())
```
and put that into the `test_that()` call directly.

hfrick (commented):
So excited to see that metric being added! 🤩 

I spotted one detail in there that's worth updating, I think, and that's the definition of which observations go into `category_1` vs `category_2`.

Other than that I left a couple of stylistic suggestions in the documentation, feel free to ignore those if you don't like them.

R/surv-brier.R
```diff
@@ -0,0 +1,166 @@
...
+                                      na_rm = TRUE,
+                                      case_weights = NULL,
+                                      ...) {
+  data <- dplyr::group_by(data, {{eval_time}})
+
+  dynamic_survival_metric_summarizer(
+    name = "brier_survival",
+    fn = brier_survival_vec,
+    data = data,
+    truth = !!enquo(truth),
```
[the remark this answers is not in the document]
EmilHvitfeldt: for now, I would just keep it consistent with the other functions

R/surv-brier.R
EmilHvitfeldt: It needed to be said, otherwise I would get back in 2 years and have no idea 😆

tests/testthat/test-surv-brier.R
EmilHvitfeldt: good catch!

R/surv-brier.R
EmilHvitfeldt: We are doing that here https://github.com/tidymodels/yardstick/blob/main/R/template.R#L329

tests/testthat/test-surv-brier.R
EmilHvitfeldt: That is a good point. I am currently following the existing codebase where these functions do more complicated things. I'll open an issue to work on that globally

R/surv-brier.R
hfrick: Agree! 😄

R/surv-brier.R
hfrick: ah, nice. then we should pass the dots from `brier_survival.data.frame()` to `dynamic_survival_metric_summarizer()` though

PR #372, opened 2023-02-24
Add dynamic_survival_metric_summarizer()

EmilHvitfeldt:
This PR is the second in a series of PRs were I chip away at https://github.com/tidymodels/yardstick/pull/347, for ease of reviewing.

This PR:

- Adds `finalize_estimator_default.Surv()`, which is mainly important, in the tests here: https://github.com/tidymodels/yardstick/compare/survival-2#diff-1d2bf29a80882aeed819f5e45a4eaab8518e43ac211ed0c2d93f53a76df4b66eR844
- Update `yardstick_remove_missing()` and `yardstick_any_missing()`
- Adds `dynamic_survival_metric_summarizer()` which is the last bit of infrastructure needed to add dynamic survival metrics.

Issue #344, opened 2022-11-26
Error in curve function if `na_rm = FALSE`


EmilHvitfeldt:
It is kind of weird that `roc_curve()` would return `NA` in the case of `na_rm = FALSE` and missing values, right?

Because we otherwise return a full tibble, so that isn't type stable.

I think this inconsistency probably already existed, but I think we should open an issue about this for all the curve functions. I imagine we should probably just error if `na_rm = FALSE` for a curve function and we actually detect missing values? That would be a nice way to take advantage of the fact that we have inlined the missing value handling into the metric function itself!

_Originally posted by @DavisVaughan in https://github.com/tidymodels/yardstick/pull/337#discussion_r1031458686_

PR #376, opened 2023-02-28
Curve survival metric summarizer

EmilHvitfeldt:
This PR is the forth in a series of PRs were I chip away at https://github.com/tidymodels/yardstick/pull/347, for ease of reviewing.

It is an almost duplicate of https://github.com/tidymodels/yardstick/pull/372, where `curve_survival_metric_summarizer()` uses `reframe()` instead of `summarize()` which is used in `dynamic_survival_metric_summarizer()`.

This PR:

- adds `curve_survival_metric_summarizer()` and associated tests


simonpcouch (approved):
On board! This looks clean.

One question: would it be advantageous for us to route `curve_survival_metric_summarizer` and `dynamic_survival_metric_summarizer` through a helper `survival_metric_summarizer` that just takes their arguments and a `summary_fn` (one of `dplyr::summarize` or `dplyr::reframe`, in order to reduce code duplication? I'm okay if the answer to this is no, for simplicity of the codebase, but it seems like there's little that's not shared between these two functions.

EmilHvitfeldt:
it is a good question! For now I think keeping them separate is best, as things in the future might pop up as difference between the two styles of metrics. One such thing is https://github.com/tidymodels/yardstick/issues/344

PR #387, opened 2023-03-14
Add concordance_survival() metric

EmilHvitfeldt:
This is another PR that extends the work done in https://github.com/tidymodels/yardstick/pull/347.

This PR aims to:

- Add `concordance_survival()`

This metric ends up being a very thin wrapper around `survival::concordance()` which is quite nice


DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Suggests:
     knitr,
     probably (>= 0.0.6),
     rmarkdown,
-    survival,
+    survival (>= 3.5-0),
```
EmilHvitfeldt, line 37: Previous versions of {survival} had a overflow issue when calling `survival::concordance()`

tests/testthat/test-surv-concordance.R
hfrick, line 1: Either we think very differently or this is an unintentional swap but I would call the object from the survival object the "expected result" 🤔

R/surv-concordance.R
```diff
@@ -0,0 +1,113 @@
...
+                                            ...) {
+  static_survival_metric_summarizer(
+    name = "concordance_survival",
+    fn = concordance_survival_vec,
+    data = data,
+    truth = !!enquo(truth),
+    estimate = !!enquo(estimate),
+    na_rm = na_rm,
+    case_weights = !!enquo(case_weights)
+  )
```
hfrick, line 82: Is the idea again to rely on `static_survival_metric_summarizer()` to check the dots for us? If so, then we should pass those on here.

R/surv-concordance.R
```diff
@@ -0,0 +1,113 @@
...
+#'   is created using [survival::Surv()].). This should be an unquoted column
+#'   name although this argument is passed by expression and supports
+#'   [quasiquotation][rlang::quasiquotation] (you can unquote column names). For
+#'   `_vec()` functions, an [survival::Surv()] object.
+#'
+#' @param estimate The column identifier for the survival probabilities. This
+#'   needs to be a numeric vector.This should be an unquoted column name
+#'   although this argument is passed by expression and supports
+#'   [quasiquotation][rlang::quasiquotation] (you can unquote column names). For
+#'   `_vec()` functions, a numeric vector.
```
hfrick, line 42: Could these be inherited from some other documentation? The Brier score, maybe, at least for `truth` and `estimate`?

hfrick (approved):
Looks good! Glad the overflow issue got resolved upstream!

R/surv-concordance.R
EmilHvitfeldt: done!

R/surv-concordance.R
EmilHvitfeldt: Good spotted!

EmilHvitfeldt: Turns out this was correct all along. It is an old design decision we are stuck in, but we need to not pass in `...` to make metric_set()` work

PR #378, opened 2023-03-01
Add `roc_survival_curve()`

EmilHvitfeldt:
This is another PR that extends the work done in https://github.com/tidymodels/yardstick/pull/347.

This PR aims to:

- Add the time-dependent ROC curve function for right-censored data with the function `roc_survival_curve()`

To be done:

- [ ] More documentation
- [ ] tests
- [ ] references (@topepo I trust you have some here)
- [x] plotting method?

PR #390, opened 2023-03-14
Add roc_auc_survival()

EmilHvitfeldt:
This PR is based on https://github.com/tidymodels/yardstick/pull/378, to supply the `roc_auc_survival()` metric.


R/surv-roc_survival_auc.R
```diff
@@ -0,0 +1,87 @@
...
+                                 eval_time,
+                                 na_rm = TRUE,
+                                 case_weights = NULL,
+                                 ...) {
+  # No checking since roc_curve_survival_vec() does checking
+  curve <- roc_curve_survival_vec(truth, estimate, censoring_weights, eval_time)
+  roc_trap_auc(curve$specificity, curve$sensitivity)
+}
+
+roc_trap_auc <- function(specificity, sensitivity) {
```
simonpcouch, line 81: Is "trap" usual verbiage here? Thumbs up if so, but I feel like "calculate" is what I'm used to hearing.

R/surv-roc_survival_auc.R
```diff
@@ -0,0 +1,87 @@
...
+  not_na <- !is.na(sensitivity) & !is.na(specificity)
+  sensitivity <- sensitivity[not_na]
+  specificity <- specificity[not_na]
+
+  auc( 1 - specificity, sensitivity)
```
simonpcouch, line 86: Feels worth noting that a different pattern exists here:

https://github.com/tidymodels/yardstick/blob/4a620f3f41289a53e7eda75ae46354c616e35755/R/prob-roc_auc.R#L212-L219

This is because `direction = "minimize"`?

simonpcouch (approved):
A bit hesitant to Approve as I'm not sure I have the expertise to give this a proper thumbs up, but everything looks fine to me!

R/surv-roc_survival_auc.R
EmilHvitfeldt: Yes, it is short for trapezoidal rule

R/surv-roc_survival_auc.R
EmilHvitfeldt: It should have been `"maximize"` 😅

simonpcouch: I'm still a bit confused here, then—if we want to maximize this metric, shouldn't we pass `specificity` rather than `1 - specificity`, as in the code linked above? Or is the thing labeled `specificity` actually `one_minus_specificity` in the first place?

Issue #348, opened 2022-12-10
Can `validate_class()` print the function that called it?


mikemahoney218:
## The problem

Right now, if you hand a yardstick function the wrong class, you get an error from `validate_class()`:

``` r
yardstick::rmse_vec(c("a", "b"), c("3", "d"))
#> Error in `validate_class()`:
#> ! `truth` should be a numeric but a character was supplied.
```

<sup>Created on 2022-12-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Similar errors come from passing the wrong type of object:

``` r
yardstick::rmse_vec(c(3, 2), matrix(2, 1))
#> Error in `validate_truth_estimate_types()`:
#> ! `estimate` should be a numeric vector, not a numeric matrix.
```

<sup>Created on 2022-12-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I think this might be friendlier if (in this situation) the package printed `Error in 'rmse_vec()'`, using the original call.

DavisVaughan:
I think this could be part of a much larger overhaul of yardstick that passes the error call argument of `abort()` through correctly, which would automatically result in the correct function name being displayed.

It could definitely be done in multiple PRs though, with maybe one master issue tracking what needs to be done

PR #405, opened 2023-04-06
don't use dplyr::summarize() in numeric_metric_summarizer()

EmilHvitfeldt:
This PR will be a small rewrite of the `*_metric_summarizer()` functions, to avoid using `summarize()` and `reframe()` since they are messing up our ability to pass the call argument around for better errors. See https://github.com/tidymodels/yardstick/issues/348


simonpcouch:
So stoked about this!

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
+  group_rows <- dplyr::group_data(data)$.rows
+  groups_data <- vctrs::vec_chop(data, indices = group_rows)
+  out <- list()
+
+  for (i in seq_along(groups_data)) {
+    .data <- groups_data[[i]]
+
+    if (!is.null(case_weights)) {
+      case_weights <- .data[[case_weights]]
+    }
```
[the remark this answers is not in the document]
DavisVaughan: If you rename the other things to `elt_` then I'd also rename this to `elt_case_weights` for consistency

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
+
+    out[[i]] <- list(
+      .metric = name,
+      .estimator = finalize_estimator(
+        .data[[truth]],
+        metric_class = name
+      ),
+      .estimate = rlang::inject(
```
DavisVaughan, line 131: Injection is a little gross, I forgot we'd have to do that, but I don't see a better way since we don't expect that the underlying vector functions capture the `...` with `enquos()` or `list2()` or something that can handle the `!!!`

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
+          case_weights = case_weights,
+          na_rm = na_rm,
+          !!!fn_options
+        )
+      )
     )
-  )
+  }
 
+  out <- dplyr::bind_rows(out)
```
DavisVaughan, line 143: When you switch to using `new_tibble()`, then I'd also switch to `vctrs::vec_rbind(!!!out)`

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
   dplyr::as_tibble(out)
```
DavisVaughan, line 144: Should be able to delete this if all of the elements of `out` were tibbles before being `vec_rbind()`-ed

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
-      .data[[truth]],
-      metric_class = .env[["name"]]
-    ),
-    .estimate = fn(
-      truth = .data[[truth]],
-      estimate = .data[[estimate]],
-      case_weights = !!case_weights,
-      na_rm = .env[["na_rm"]],
-      !!!fn_options
+  group_rows <- dplyr::group_data(data)$.rows
```
DavisVaughan, line 114:
```suggestion
  group_rows <- dplyr::group_rows(data)
```

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
+  groups_data <- vctrs::vec_chop(data, indices = group_rows)
```
DavisVaughan, line 115: One naming scheme you could use is to call this `groups <- ` and the in the loop at each iteration you do

```r
group <- groups[[i]]
```

I like that better than the `elt` idea. I think I typically only do `elt` if the thing I'm iterating over isn't a plural word (i.e. `groups` is plural so you don't need `elt` since you can use `group`)

Then you'd have `group_truth`, `group_estimate`, and `group_case_weights`

DavisVaughan (approved):
Looks pretty good. I think eventually we will wrap the call to `fn()` in `withCallingHandlers()` to chain any errors in the underlying vector function to the data frame function, but that can come later

R/template.R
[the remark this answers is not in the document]
EmilHvitfeldt: fixed in [`6e3e1d8` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/6e3e1d84ff6e59e57ab46f55988a5341f19f9022)

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
+    .data <- groups_data[[i]]
+
+    if (!is.null(case_weights)) {
+      case_weights <- .data[[case_weights]]
+    }
+
+    out[[i]] <- list(
+      .metric = name,
+      .estimator = finalize_estimator(
+        .data[[truth]],
```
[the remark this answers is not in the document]
EmilHvitfeldt: finalized in [`f272a12` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/f272a12af1d6b5ccf4b16af6598af88557737cac)

R/template.R
EmilHvitfeldt: done in [`f272a12` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/f272a12af1d6b5ccf4b16af6598af88557737cac)

R/template.R
EmilHvitfeldt: [`949ff01` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/949ff01b112e6827ac49036a2a20bf0ae24585ab)

R/template.R
EmilHvitfeldt: https://github.com/tidymodels/yardstick/pull/405/commits/949ff01b112e6827ac49036a2a20bf0ae24585ab

R/template.R
EmilHvitfeldt: I know, I was happy that at least it worked 😅

R/template.R
```diff
@@ -100,33 +100,47 @@ numeric_metric_summarizer <- function(name,
...
+  out <- list()
+
+  for (i in seq_along(groups_data)) {
+    .data <- groups_data[[i]]
+
+    if (!is.null(case_weights)) {
+      case_weights <- .data[[case_weights]]
+    }
+
+    out[[i]] <- list(
```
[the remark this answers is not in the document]
EmilHvitfeldt: done in [`55be7c6` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/55be7c6522de30c81a65a9116038a2e74078e80a)

R/template.R
```diff
@@ -288,40 +348,58 @@ curve_metric_summarizer <- function(name,
...
+            !!! spliceable_argument(event_level, "event_level"),
+            !!! fn_options
+          )
+        )
+      )
     )
-  )
 
-  dplyr::as_tibble(out)
+    out[[i]] <- tibble::as_tibble(elt_out)
```
DavisVaughan, line 399:
```suggestion
    out[[i]] <- tibble::new_tibble(elt_out)
```

typo or necessary?

R/template.R
```diff
@@ -350,38 +428,62 @@ dynamic_survival_metric_summarizer <- function(name,
...
+            case_weights = group_case_weights,
+            na_rm = na_rm,
+            !!!fn_options
+          )
+        )
+      )
     )
-  )
+
+    out[[i]] <- tibble::as_tibble(elt_out)
```
DavisVaughan, line 476:
```suggestion
    out[[i]] <- tibble::new_tibble(elt_out)
```

R/template.R
```diff
@@ -468,34 +594,55 @@ curve_survival_metric_summarizer <- function(name,
...
+            na_rm = na_rm,
+            !!! fn_options
+          )
+        )
+      )
     )
-  )
 
-  dplyr::as_tibble(out)
+    out[[i]] <- tibble::as_tibble(elt_out)
```
DavisVaughan, line 642:
```suggestion
    out[[i]] <- tibble::new_tibble(elt_out)
```

R/template.R
EmilHvitfeldt: Necessary (or maybe bug in `new_tibble()`)

``` r
elt_out <- list(
  .metric = "roc_curve",
  .estimator = "binary",
  .estimate = tibble::tibble(
    .threshold = c(-Inf, 0.1, 0.5, Inf),
    specificity = c(0, 0.25, 0.7, 1),
    sensitivity = c(1, 0.8, 0.2, 0)
  )
)

tibble::as_tibble(elt_out)
#> # A tibble: 4 × 3
#>   .metric   .estimator .estimate$.threshold $specificity $sensitivity
#>   <chr>     <chr>                     <dbl>        <dbl>        <dbl>
#> 1 roc_curve binary                   -Inf           0             1  
#> 2 roc_curve binary                      0.1         0.25          0.8
#> 3 roc_curve binary                      0.5         0.7           0.2
#> 4 roc_curve binary                    Inf           1             0

tibble::new_tibble(elt_out)
#> Error in `[<-`:
#> ! Assigned data `map(.subset(x, unname), vectbl_set_names, NULL)` must
#>   be compatible with existing data.
#> ✖ Existing data has 1 row.
#> ✖ Element 3 of assigned data has 4 rows.
#> ℹ Row updates require a list value. Do you need `list()` or `as.list()`?
#> Caused by error in `vectbl_recycle_rhs_rows()`:
#> ! Can't recycle input of size 4 to size 1.
```

R/template.R
```diff
@@ -100,34 +100,55 @@ numeric_metric_summarizer <- function(name,
...
-      metric_class = .env[["name"]]
-    ),
-    .estimate = fn(
-      truth = .data[[truth]],
-      estimate = .data[[estimate]],
-      case_weights = !!case_weights,
-      na_rm = .env[["na_rm"]],
-      !!!fn_options
+  group_rows <- dplyr::group_rows(data)
+  groups <- vctrs::vec_chop(data, indices = group_rows)
```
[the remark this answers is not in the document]
EmilHvitfeldt: I kept the groups to be able to slap them back on the output https://github.com/tidymodels/yardstick/pull/405/files#diff-1d635b2cef66759975cc44556f49c9796d21ccb3e7e2ba74dd79bfe9f9a195a3R129

R/template.R
DavisVaughan: Oh it is because `new_tibble()` assumes everything is the same length, and you need recycling to kick in first for the curve functions. You could use `new_tibble()` if you add this in before it:

```r
elt_out <- vctrs::vec_recycle_common(!!!elt_out)
```

It makes sense to do that for the curve summarizers. I'm not sure if it makes sense for `numeric_metric_summarizer()` and the other non-curve summarizers, because I think we expect that each group should return exactly 1 row there?

It is kind of nice that this forces an error if that is _not_ the case

EmilHvitfeldt: dealt with in [tidymodels/yardstick@`0c17e1b` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/0c17e1bbee9c66272c3956b95897ccb5c984be7f)

R/template.R
EmilHvitfeldt: dealt with in [tidymodels/yardstick@`0c17e1b` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/0c17e1bbee9c66272c3956b95897ccb5c984be7f)

R/template.R
EmilHvitfeldt: dealt with in [tidymodels/yardstick@`0c17e1b` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/0c17e1bbee9c66272c3956b95897ccb5c984be7f)

R/template.R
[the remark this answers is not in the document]
DavisVaughan: and

```r
group_keys <- vctrs::vec_rep_each(group_keys, times = list_sizes(out))

```

is only necessary for the curve ones, where we do `vec_recycle_common()`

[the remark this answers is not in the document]
EmilHvitfeldt: done here [`07ea277` (#405)](https://github.com/tidymodels/yardstick/pull/405/commits/07ea277cf03411ad4e05e8f7ce9d80e5ea84c281)

DavisVaughan (approved):

PR #420, opened 2023-04-13
Don't namespace functions from rlang or vctrs

EmilHvitfeldt:
Both {rlang} and {vctrs} are imported. So I'm going to make this change to remove some visual clutter in the codebase


DavisVaughan (approved):
Definitely agree!

PR #422, opened 2023-04-18
fully support class_pred input in metrics

### LINKED ISSUE #341: `<class_pred>` objects don't work everywhere
``` r
library(yardstick)
#> For binary classification, the first factor level is assumed to be the event.
#> Use the argument `event_level = "second"` to alter this as needed.

df <- tibble::tibble(
  truth = probably::class_pred(factor(c("a", "b", "a", "c"))), 
  estimate = factor(c("a", "c", "a", "b"))
)

accuracy(df, truth, estimate)
#> Error in `dplyr::summarise()`:
#> ℹ In argument: `.estimate = fn(truth = truth, estimate = estimate, na_rm
#>   = na_rm)`.
#> Caused by error in `binary_checks()`:
#> ! `estimator` is binary, only two class `truth` factors are allowed. A factor with 3 levels was provided.

#> Backtrace:
#>      ▆
#>   1. ├─yardstick::accuracy(df, truth, estimate)
#>   2. ├─yardstick:::accuracy.data.frame(df, truth, estimate)
#>   3. │ └─yardstick::class_metric_summarizer(...)
#>   4. │   ├─dplyr::summarise(...)
#>   5. │   └─dplyr:::summarise.data.frame(...)
#>   6. │     └─dplyr:::summarise_cols(.data, dplyr_quosures(...))
#>   7. │       ├─base::withCallingHandlers(...)
#>   8. │       └─dplyr:::map(quosures, summarise_eval_one, mask = mask)
#>   9. │         └─base::lapply(.x, .f, ...)
#>  10. │           └─dplyr (local) FUN(X[[i]], ...)
#>  11. │             └─mask$eval_all_summarise(quo)
#>  12. │               └─dplyr (local) eval()
#>  13. └─yardstick (local) fn(truth = truth, estimate = estimate, na_rm = na_rm)
#>  14.   └─yardstick::metric_vec_template(...)
#>  15.     └─yardstick:::validate_truth_estimate_checks(...)
#>  16.       ├─yardstick:::validate_truth_estimate_types(truth, estimate, estimator)
#>  17.       └─yardstick:::validate_truth_estimate_types.factor(...)
#>  18.         ├─yardstick:::binary_checks(truth, estimate)
#>  19.         └─yardstick:::binary_checks.factor(truth, estimate)
#>  20.           └─rlang::abort(...)
```

I think `finalize_estimator()` is getting a `<class_pred>`, so its internal dispatch doesn't treat it like a factor. It might need to convert to a factor early before calling `finalize_estimator_internal()`

EmilHvitfeldt:
This PR aims to close #341 

It
- uses `as_factor_from_class_pred()` inside each `_vec()` function. (i would have liked to have done it in the metric summarizers but then `_vec()` would work.
- Adds tests for each metric that supports class_pred
- Removed now redudant calls to `as_factor_from_class_pred()` https://github.com/tidymodels/yardstick/commit/1a1ce89b2718bf696a8cddd814cda12ccbb1dc9e

``` r
library(yardstick)

df <- tibble::tibble(
  truth = probably::class_pred(factor(c("a", "b", "a", "c"))), 
  estimate = factor(c("a", "c", "a", "b"))
)

accuracy(df, truth, estimate)
#> # A tibble: 1 × 3
#>   .metric  .estimator .estimate
#>   <chr>    <chr>          <dbl>
#> 1 accuracy binary           0.5

accuracy_vec(df$truth, df$estimate)
#> [1] 0.5
```

### DIFF
--- a/R/class-accuracy.R
+++ b/R/class-accuracy.R
@@ -82,6 +82,13 @@ accuracy.matrix <- function(data, ...) {
 #' @export
 #' @rdname accuracy
 accuracy_vec <- function(truth, estimate, na_rm = TRUE, case_weights = NULL, ...) {
+  if (is_class_pred(truth)) {
+    truth <- as_factor_from_class_pred(truth)
+  }
+  if (is_class_pred(estimate)) {
+    estimate <- as_factor_from_class_pred(estimate)
+  }
+
   estimator <- finalize_estimator(truth, metric_class = "accuracy")
 
   check_class_metric(truth, estimate, case_weights, estimator)
--- a/R/class-accuracy.R
+++ b/R/class-accuracy.R
@@ -83,10 +83,7 @@ accuracy.matrix <- function(data, ...) {
 #' @rdname accuracy
 accuracy_vec <- function(truth, estimate, na_rm = TRUE, case_weights = NULL, ...) {
   abort_if_class_pred(truth)
-
-  if (is_class_pred(estimate)) {
-    estimate <- as_factor_from_class_pred(estimate)
-  }
+  estimate <- as_factor_from_class_pred(estimate)
 
   estimator <- finalize_estimator(truth, metric_class = "accuracy")
   check_class_metric(truth, estimate, case_weights, estimator)
--- a/R/class-accuracy.R
+++ b/R/class-accuracy.R
@@ -82,15 +82,13 @@ accuracy.matrix <- function(data, ...) {
 #' @export
 #' @rdname accuracy
 accuracy_vec <- function(truth, estimate, na_rm = TRUE, case_weights = NULL, ...) {
-  if (is_class_pred(truth)) {
-    truth <- as_factor_from_class_pred(truth)
-  }
+  abort_if_class_pred(truth)
+
   if (is_class_pred(estimate)) {
     estimate <- as_factor_from_class_pred(estimate)
   }
 
   estimator <- finalize_estimator(truth, metric_class = "accuracy")
-
   check_class_metric(truth, estimate, case_weights, estimator)
 
   if (na_rm) {
--- a/R/class-bal_accuracy.R
+++ b/R/class-bal_accuracy.R
@@ -82,10 +82,7 @@ bal_accuracy_vec <- function(truth,
                              event_level = yardstick_event_level(),
                              ...) {
   abort_if_class_pred(truth)
-
-  if (is_class_pred(estimate)) {
-    estimate <- as_factor_from_class_pred(estimate)
-  }
+  estimate <- as_factor_from_class_pred(estimate)
 
   estimator <- finalize_estimator(truth, estimator)
 
--- a/R/class-bal_accuracy.R
+++ b/R/class-bal_accuracy.R
@@ -81,6 +81,13 @@ bal_accuracy_vec <- function(truth,
                              case_weights = NULL,
                              event_level = yardstick_event_level(),
                              ...) {
+  if (is_class_pred(truth)) {
+    truth <- as_factor_from_class_pred(truth)
+  }
+  if (is_class_pred(estimate)) {
+    estimate <- as_factor_from_class_pred(estimate)
+  }
+
   estimator <- finalize_estimator(truth, estimator)
 
   check_class_metric(truth, estimate, case_weights, estimator)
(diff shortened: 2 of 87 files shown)


R/class-accuracy.R
```diff
@@ -82,6 +82,13 @@ accuracy.matrix <- function(data, ...) {
 #' @export
 #' @rdname accuracy
 accuracy_vec <- function(truth, estimate, na_rm = TRUE, case_weights = NULL, ...) {
+  if (is_class_pred(truth)) {
```
topepo, line 85: You don't need to check the observed data, just the predicted classes.

R/class-bal_accuracy.R
```diff
@@ -81,6 +81,13 @@ bal_accuracy_vec <- function(truth,
                              case_weights = NULL,
                              event_level = yardstick_event_level(),
                              ...) {
+  if (is_class_pred(truth)) {
```
topepo, line 84: From [R4DS](https://r4ds.had.co.nz/functions.html#when-should-you-write-a-function)

> ... the “do not repeat yourself” (or DRY) principle. The more repetition you have in your code, the more places you need to remember to update when things change (and they always do!), and the more likely you are to create bugs over time.

topepo: Is there a constructor or some other commonly used function to put this in? We'll have to remember to add this to every classification metric that uses hard class predictions.

Maybe we should add one that is automatically invoked somewhere?

R/prob-brier.R
```diff
@@ -83,6 +83,10 @@ brier_class_vec <- function(truth,
                             na_rm = TRUE,
                             case_weights = NULL,
                             ...) {
+  if (is_class_pred(truth)) {
```
topepo, line 86: Does this metric use the hard class predictions?

topepo: Ok, I see that a lot of other probability metrics have this. I don't think that they should/do use this column.

topepo (changes requested):

R/class-bal_accuracy.R
EmilHvitfeldt: > Is there a constructor or some other commonly used function to put this in? We'll have to remember to add this to every classification metric that uses hard class predictions.

Short answer: no.

at first I wanted to put this in the metric summarizers such as `class_metric_summarizer()`, to avoid copying all this code around. However, when I did that, all the `*_vec()` functions didn't work because they don't call the metric summarizers. 

From https://github.com/tidymodels/yardstick/issues/341 we have that `finalize_estimator()` is the typical culprit in errors, and since that is the first thing called in the `*_vec()` functions I sadly had to put that code there.

This is of cause predicated on the assumption that we want both `accuracy()` and `accuracy_vec()` to be able to deal with `class_pred` which seems reasonable.

R/class-accuracy.R
EmilHvitfeldt: Sounds good, I'll change that.

Just for documentation sake, we used to "support" both `truth` and `estimate` as `class_pred` objects
https://github.com/tidymodels/yardstick/blob/d4a9d7af3f44f8dbc8a78827e1ed79b9e46c895d/R/metrics.R#L85-L90

topepo: It is inappropriate for `truth` to be a `class_pred` object, so we should error in that case.

R/class-bal_accuracy.R
topepo: ok., I would at least make it a function that gets invoked a lot.

R/class-accuracy.R
EmilHvitfeldt: Done in [tidymodels/yardstick@`0352475` (#422)](https://github.com/tidymodels/yardstick/pull/422/commits/0352475d7273b4017465db775728043ceca492d0)

R/prob-brier.R
EmilHvitfeldt: Done in [tidymodels/yardstick@`0352475` (#422)](https://github.com/tidymodels/yardstick/pull/422/commits/0352475d7273b4017465db775728043ceca492d0)

R/class-bal_accuracy.R
EmilHvitfeldt: done here [tidymodels/yardstick@`90fa6af` (#422)](https://github.com/tidymodels/yardstick/pull/422/commits/90fa6af29bb5aa3995dd2a5c4bca49991a838241)

PR #428, opened 2023-05-01
cache known selections while tuning

simonpcouch:
In tune, `yardstick_eval_select()` is called `n_resamples * n_grid_points * n_metrics` times for each column selection input. Within a tuning process, the `expr` passed to `eval_select()` uniquely identifies the selection result. We can thus cache those known selections after they're evaluated once and then retrieve them for the remainder of the tuning process. 

In context, this can have a substantial impact on time to evaluate against resamples, especially with quick model fits and many resamples / grid points. With this PR:

``` r
bench::mark(
  new = 
    fit_resamples(
      logistic_reg(),
      Class ~ .,
      bootstraps(two_class_dat, 1000)
    )
)
#> # A tibble: 1 × 13
#>   expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 new           10.7s  10.7s    0.0935    2.19GB     3.18     1    34
#> # ℹ 5 more variables: total_time <bch:tm>, result <list>,
#> #   memory <list>, time <list>, gc <list>
```

With `main` dev:

``` r
#> # A tibble: 1 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 old           12.4s    12.4s    0.0805     2.2GB     4.59     1    57
#> # ℹ 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#> #   time <list>, gc <list>
```

When this machinery isn't kicked in, i.e. when yardstick functions are called outside of a tune function, the overhead is just the time that `in_tuning_env()` takes (after the first call to a yardstick function, per session):

``` r
bench::mark(
    in_tuning_env()
)
#> # A tibble: 1 × 13
#>   expression           min  median `itr/sec` mem_alloc `gc/sec` n_itr
#>   <bch:expr>      <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl> <int>
#> 1 in_tuning_env()   1.97µs  2.54µs   334133.        0B        0 10000
#> # ℹ 6 more variables: n_gc <dbl>, total_time <bch:tm>,
#> #   result <list>, memory <list>, time <list>, gc <list>
```

PR 1/2: further testing heading to extratests in a moment!

### DIFF
--- a/R/template.R
+++ b/R/template.R
@@ -759,6 +759,14 @@ yardstick_eval_select <- function(expr,
                                   error_call = caller_env()) {
   check_dots_empty(call = error_call)
 
+  if (!quo_is_missing(expr) && inherits(quo_get_expr(expr), "name")) {
+    expr_name <- as_name(expr)
+
+    if (is_known_selection(expr_name)) {
+      return(get_known_selection(expr_name))
+    }
+  }
+
   out <- tidyselect::eval_select(
     expr = expr,
     data = data,
@@ -774,12 +782,24 @@ yardstick_eval_select <- function(expr,
     abort(message, call = error_call)
   }
 
+  set_known_selection(as_name(expr), out)
+
   out
 }
 
 yardstick_eval_select_dots <- function(...,
                                        data,
                                        error_call = caller_env()) {
+  expr <- quo(!!substitute(...))
+
+  if (!quo_is_missing(expr) && inherits(quo_get_expr(expr), c("name", "call"))) {
+    expr_label <- as_label(expr)
+
+    if (is_known_selection(expr_label)) {
+      return(get_known_selection(expr_label))
+    }
+  }
+
   out <- tidyselect::eval_select(
     expr = expr(c(...)),
     data = data,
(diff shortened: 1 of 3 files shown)


R/template.R
```diff
@@ -791,5 +811,78 @@ yardstick_eval_select_dots <- function(...,
...
+  tune_env$known_selections[[expr_name]] <- out
+
+  invisible()
+}
+
+init_known_selection <- function(tune_env) {
+  withr::defer(
+    env_bind(tune_env, known_selections = NULL),
+    envir = tune_env$progress_env
+  )
```
simonpcouch, line 856: See [these lines in tune](https://github.com/tidymodels/tune/blob/28a459f9f581bf35d5c28af67866588f71056b36/R/logging.R#L110) for context on this environment--`progress_env` persists while a given tune function, like `fit_resamples()` or `tune_bayes()`, runs. When that function returns, this code wipes the known selections so that we don't re-use them across different calls to tuning functions.

R/template.R
```diff
@@ -791,5 +811,78 @@ yardstick_eval_select_dots <- function(...,
...
+    return(invisible())
+  }
+
+  tune_env <- ns_env("tune")$tune_env
+
+  if (is.null(tune_env$known_selections)) {
+    init_known_selection(tune_env)
+  }
+
+  tune_env$known_selections[[expr_name]] <- out
```
simonpcouch, line 847: Note:

``` r
boop <- NULL
boop[["a"]] <- 1
boop
#> $a
#> [1] 1
```

<sup>Created on 2023-05-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

So no need to initialize `known_selections()`.

tests/testthat/test-template.R
```diff
@@ -1719,3 +1719,12 @@ test_that("curve_survival_metric_summarizer() handles column name collisions", {
 
   expect_identical(roc_survival_curve_res, roc_survival_curve_exp)
 })
+
+test_that("known selections don't affect selection without tune", {
+  # yardstick's CI reliably does not have tune installed
+  skip_if(rlang::is_installed("tune"), "tune is installed")
```
EmilHvitfeldt, line 1725: Should use `testthat::skip_if_not_installed()`

R/template.R
```diff
@@ -759,6 +759,14 @@ yardstick_eval_select <- function(expr,
                                   error_call = caller_env()) {
   check_dots_empty(call = error_call)
 
+  if (!quo_is_missing(expr) && inherits(quo_get_expr(expr), "name")) {
+    expr_name <- as_name(expr)
```
EmilHvitfeldt, line 763: not a big deal, but if you refactor out this to the line before then you can say 

```r
  set_known_selection(expr_name, out)
```

instead of

```r
  set_known_selection(as_name(expr), out)
```

R/template.R
```diff
@@ -774,12 +782,24 @@ yardstick_eval_select <- function(expr,
...
   out
 }
 
 yardstick_eval_select_dots <- function(...,
                                        data,
                                        error_call = caller_env()) {
+  expr <- quo(!!substitute(...))
+
+  if (!quo_is_missing(expr) && inherits(quo_get_expr(expr), c("name", "call"))) {
+    expr_label <- as_label(expr)
```
EmilHvitfeldt, line 796: Same comment as before

EmilHvitfeldt (approved):
looks good, very exciting change. couple of notes

R/template.R
simonpcouch: Yes, but `as_name()` expr errors if either of the above conditions aren't true.

tests/testthat/test-template.R
EmilHvitfeldt: you are right, please ignore! :D

EmilHvitfeldt:
One thing we talked about is the idea that {yardstick} shouldn't know about {tune}. And this would be in violation to that goal.

I'm going to allow this change, as we can rip out this code if {tidyselect} becomes fast enough for this to be unimportant.

Issue #421, opened 2023-04-16
Group_by calculate metric


SHo-JANG:
Here's what I want to do specifically.
For example, let's say I have monthly trading data for all tickers in the stock market.
I want to be able to sort the predicted returns for all stocks by year and month.
Then, I want to calculate a statistic for only the top 10% of stocks by predicted return for each month of the year.
The specific metric is up to you, but you can calculate the RMSE of the top 10% .
The goal is to tune the hyperparameters so that the actual returns of the top 10% predicted stocks are higher.

In other words, I want to find a hyperparameter that tends to get the top 10% right, even if it gets the bottom 90% wrong, rather than getting the whole universe right.

I tried to define custom_metric well, but it was limited. I wanted to put **`group_by(yearmonth)`** in the process, but I didn't really know how to do it. 
so I made a makeshift code. 

```r
# customize_metric --------------------------------------------------------
# irr = The return of a portfolio of stocks with a predicted top 10% return, calculated monthly.
# return_pct = monthly cumulative return ratio. ex. 5.1~ 5.31 's cumulative return ratio.
irr_vec <- function(truth,
                    estimate,
                    n_tiles =10,
                    purpose_tile = 10, #predicted top 10% return =10 , bottom 10% = 1
                    na_rm = TRUE,
                    case_weights = NULL,
                    ...) {
  
  
  
  irr_impl <- function(truth, estimate,..., case_weights = NULL) {
    
    
    fold_index <<- NULL
    for( i in 1:dim(valid_years_splited)[1]){
      if(length(truth) == nrow(valid_years_splited$data[[i]]) ){
        fold_index <<-i
      } 
    }
    
    
    valid_years_splited$data[[fold_index]] |> 
      mutate(estimate = estimate) |> 
      group_by(yearmonth) |> 
      mutate(top_n_pct = ntile(estimate,10)) |> 
      filter(top_n_pct == purpose_tile) |> 
  #mean_y := portpolio which is composed by predicted return top10% 
      summarise(mean_y = mean(return_pct)) |> 
 #irr := portpolio 1year cumulative return ratio
      mutate(irr =cumprod(mean_y/100+1) )  |> slice_tail(n = 1) |> pull(irr) ->irr
    
    
    #If another folder has same length each other, this code is unusable .
    
# cross -validation summarize by mean default. So I can calculate geometric mean by log( )
    return(log(irr))      
    
    
  }
  
  metric_vec_template(
    metric_impl = irr_impl,
[...]

Issue #371, opened 2023-02-23
Support for grouped custom metrics in workflows


simdadim:
I've working on a tabular dataset where multiple rows are linked forming an "session". Within the session, one of the rows are our target action, and the goal for the model is to try to find this, and bump this row as far up within the session as possible (predict this row with higher probability than the other rows in the same session). 

I've modeled this using a binary classifier. To evaluate the performance, I want to se how much up/down the target row was moved in the correct direction if I sort the rows within the session by each rows predicted probability of being the correct one. Stated otherwise - I want the most likely row within each session to have the highest probability within this session.

The formula is quite easy, its just group by session_id, identify the target row in each session, and evaluate the relative distance it has moved if I sort on the new predicted probability score.

But, I struggle to create a custom yardstick-metric to calculate this, since the grouping of the data is not passed on to my custom evaluating function. I've tried different approaches, but from what I can see, the problem is that the `predict_model`-function is dropping the grouping of the dataframe.

Is it possible to keep the grouping of a DF in `predict_model`, and include the grouping variable? To my understanding, this would make it possible to develop custom metrics that accounts for grouped data. I imagine this means that the `.predictions` column in the `resample_results` dataframe would also keep the groups.

EmilHvitfeldt:
Hello @simdadim 👋 

This is a good idea, we are currently thinking about how to best handle these types of metrics. We want to make sure our approach is sound and general enough to do everything we want.

I'm gonna keep this issue up to remind us of your request, but it will take a little while before we get to working on this problem.

Issue #176, opened 2020-07-10
Fairness metrics


juliasilge:
Let's look into supporting metrics for measuring fairness for models.

- There is an [existing R package that includes such metrics](https://github.com/kozodoi/fairness).

- There are [fairness metrics in an open source library](https://github.com/IBM/AIF360) (R and Python) from IBM.

- There are definitions of [several metrics on Wikipedia](https://en.wikipedia.org/wiki/Fairness_(machine_learning)#Metrics%5B6%5D).

juliasilge:
h2o uses some different names for some of the same metrics:

- Adverse Impact Ratio (classification)
- Standardized Mean Difference (regression)

apreshill:
Adding fairness indicators for Tensorflow: https://github.com/tensorflow/fairness-indicators

EmilHvitfeldt:
It might be a good idea to have some biased datasets in modeldata for examples.

One commonly used dataset is the [Boston housing dataset](https://www.kaggle.com/c/boston-housing) which includes problematic variables such as 

- black: 1000(Bk - 0.63)^2 where Bk is the proportion of blacks by town.
- lstat: lower status of the population (percent).

I don't know how I feel about it ethically, but with enough documentation, it might be okay.

EmilHvitfeldt:
I would like it if there would be some indication on the pkgdown reference page that a metric is fairness related.

topepo:
`fairness` has different functions for different metrics. We should try to have a more general method that takes a metric set or some other metric-related argument.

juliasilge:
The [fairmodels](https://modeloriented.github.io/fairmodels/) package computes fairness metrics based on confusion matrices and a DALEX explainer.

skeydan:
I would say that both AI Fairness 360 and the `fairness` package (both mentioned by @juliasilge in the first comment) look very good. (By the way, they include the standard fairness datasets as well, so probably no need to have one's own?).

To me then this issue seems to be first and foremost about integration into `tidymodels`, right? (Meaning, it's a software engineering question.) 
However, there is also the question of what the metrics actually measure / hope to measure, and even though AI F 360 has good docs on that too (and a nice demo), I think one could make more explicit the underlying assumptions ...

I'll open a broader issue in a repo more appropriate for this, and hope to be linking that back here later.

@topepo @tracykteal

topepo:
We'll work on another package for this. It will use yardstick to compute individual metrics but be another thing.

PR #434, opened 2023-05-15
add fairness metrics

simonpcouch:
First, an example with tune:

``` r
library(tidymodels)

# check out gss data from the infer package
gss
#> # A tibble: 500 × 11
#>     year   age sex    college   partyid hompop hours income class finrela weight
#>    <dbl> <dbl> <fct>  <fct>     <fct>    <dbl> <dbl> <ord>  <fct> <fct>    <dbl>
#>  1  2014    36 male   degree    ind          3    50 $2500… midd… below …  0.896
#>  2  1994    34 female no degree rep          4    31 $2000… work… below …  1.08 
#>  3  1998    24 male   degree    ind          1    40 $2500… work… below …  0.550
#>  4  1996    42 male   no degree ind          4    40 $2500… work… above …  1.09 
#>  5  1994    31 male   degree    rep          2    40 $2500… midd… above …  1.08 
#>  6  1996    32 female no degree rep          4    53 $2500… midd… average  1.09 
#>  7  1990    48 female no degree dem          2    32 $2500… work… below …  1.06 
#>  8  2016    36 female degree    ind          1    20 $2500… midd… above …  0.478
#>  9  2000    30 female degree    rep          5    40 $2500… midd… average  1.10 
#> 10  1998    33 female no degree dem          2    40 $1500… work… far be…  0.550
#> # ℹ 490 more rows

# tune an xgboost model on college completion
res <-
  tune_grid(
    boost_tree("classification", trees = tune(), min_n = tune()),
    college ~ age + income,
    vfold_cv(gss, v = 10, repeats = 2),
    metrics = metric_set(roc_auc, demographic_parity(sex))
  )

collect_metrics(res)
#> # A tibble: 20 × 9
#>    trees min_n .metric            .estimator .by     mean     n std_err .config 
#>    <int> <int> <chr>              <chr>      <chr>  <dbl> <int>   <dbl> <chr>   
#>  1  1028     3 demographic_parity binary     sex   0.105     20 0.0160  Preproc…
#>  2  1028     3 roc_auc            binary     <NA>  0.611     20 0.0238  Preproc…
#>  3  1767     6 demographic_parity binary     sex   0.109     20 0.0161  Preproc…
#>  4  1767     6 roc_auc            binary     <NA>  0.636     20 0.0242  Preproc…
#>  5   525    12 demographic_parity binary     sex   0.107     20 0.0146  Preproc…
#>  6   525    12 roc_auc            binary     <NA>  0.643     20 0.0225  Preproc…
#>  7   998    17 demographic_parity binary     sex   0.0689    20 0.0128  Preproc…
#>  8   998    17 roc_auc            binary     <NA>  0.632     20 0.0215  Preproc…
#>  9   640    20 demographic_parity binary     sex   0.0669    20 0.00920 Preproc…
#> 10   640    20 roc_auc            binary     <NA>  0.626     20 0.0208  Preproc…
#> 11    38    22 demographic_parity binary     sex   0.0683    20 0.0160  Preproc…
#> 12    38    22 roc_auc            binary     <NA>  0.666     20 0.0207  Preproc…
#> 13  1372    25 demographic_parity binary     sex   0.0727    20 0.0138  Preproc…
#> 14  1372    25 roc_auc            binary     <NA>  0.621     20 0.0219  Preproc…
#> 15  1828    30 demographic_parity binary     sex   0.0527    20 0.0122  Preproc…
#> 16  1828    30 roc_auc            binary     <NA>  0.616     20 0.0177  Preproc…
#> 17  1509    33 demographic_parity binary     sex   0.0544    20 0.00968 Preproc…
#> 18  1509    33 roc_auc            binary     <NA>  0.614     20 0.0174  Preproc…
#> 19   252    39 demographic_parity binary     sex   0.0370    20 0.00633 Preproc…
#> 20   252    39 roc_auc            binary     <NA>  0.633     20 0.0125  Preproc…

autoplot(res)
```

This PR implements a fairness metric constructor, `fairness_metric()`, as well as 3 outputs of it giving canonical fairness metrics. One of them is `demographic_parity()`:

``` r
# quick data setup: 
gss$college_pred <- gss$college
gss$college_pred[sample(1:nrow(gss), 10)] <- "degree"

dp_sex <- demographic_parity(sex)

# demographic_parity(by) output is just a yardstick metric:
class(dp_sex)
#> [1] "class_metric" "metric"       "function"

# user interacts with it like any other:
dp_sex(gss, truth = college, estimate = college_pred)
#> # A tibble: 1 × 4
#>   .metric            .by   .estimator .estimate
#>   <chr>              <chr> <chr>          <dbl>
#> 1 demographic_parity sex   binary        0.0135
```

Under the hood, the 3 new metrics are “group-aware”, associated with a “by” data-column that, given some dataset, generates pre-existing yardstick metrics by group and then summarizes across groups back to 1 number, the usual level of observation.

They’re created with `fairness_metric()`—the actual definition of `demographic_parity()` looks like:

``` r
diff_range <- function(x, ...) {
  diff(range(x$.estimate))
}

demographic_parity <-
  fairness_metric(
    .fn = detection_prevalence,
    .name = "demographic_parity",
    .post = diff_range
  )
```

The idea here is:
1) Make the interface feel as tidymodels-idiomatic as possible—existing idioms “just work”, and
2) Create a minimal set of canonical metrics using flexible user-facing tooling; we want to encourage users (in further vignettes + docs) to make use of `fairness_metric()` to create fairness metrics with their modeling context in mind.

This is already a large PR, though there’s a lot more to do here. I felt that this was enough of a start to give a solid roadmap if we were to move ahead with this approach to assessment.

Related to \#176, \#371, \#421. tune PR following up here coming in a moment. :)

Reviewing commit-by-commit should be easier than via the whole PR!


juliasilge (commented):
My opinion is that this approach (with `.fn` and `.post`) looks excellent. We've got the basic fairness metrics included, but the ability to make new ones is fairly straightforward as well. FWIW here it feels quite a bit easier to make a new fairness metric function than to make a [custom yardstick metric](https://www.tidymodels.org/learn/develop/metrics/).

Given that we are doing a lot of function factory type of stuff, it might be nice to give folks a little more help in learning/dealing with the inputs and outputs. 
- Maybe a link to here in the docs for the current `fairness_metric()` (plus use the phrase "function factory"): https://adv-r.hadley.nz/function-factories.html
- I think a print method for the `metric` functions, if people are going to be creating them and dealing with them:

``` r
yardstick::sens
#> function (data, ...) 
#> {
#>     UseMethod("sens")
#> }
#> <bytecode: 0x117ee7e30>
#> <environment: namespace:yardstick>
#> attr(,"direction")
#> [1] "maximize"
#> attr(,"class")
#> [1] "class_metric" "metric"       "function"

## just an idea:
format.metric <- function(x, ...) {
    first_class <- class(x)[[1]]
    cli::cli_format_method({
        cli::cli_h3("A {.cls {first_class}} function to {attr(x, 'direction')} metrics")
    })
}

print.metric <- function(x, ...) {
    cat(format(x), sep = "\n")
    invisible(x)
}

yardstick::sens
#> 
#> ── A <class_metric> function to maximize metrics
```

<sup>Created on 2023-05-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/fair-aaa.R
```diff
@@ -0,0 +1,137 @@
...
+            call = rlang::call2("fairness_metric")
+          )
+        }
+
+        if (length(gp_vars) > 0) {
+          res <- dplyr::group_by(res, !!!dplyr::groups(data), .add = FALSE)
+        }
+
+        res <-
+          dplyr::summarize(
```
EmilHvitfeldt, line 109: This might take a little time to convert, but I think we should avoid calling `dplyr::summarize()` directly, and instead use loop with the dplyr group helpers. Similar to how it was done in https://github.com/tidymodels/yardstick/pull/405. I'm imagining that it would make the errors better

EmilHvitfeldt: I'm putting it here, since this issue should be resolved first. The current implementation doesn't give good errors 😞  I'm very positive we can make it better!

``` r
library(yardstick)

data(hpc_cv)

equalized_odds_resample <- equalized_odds(Resample)

hpc_cv |>
  equalized_odds_resample(truth = VF, estimate = pred)
#> Error in `metric_set()`:
#> ! Failed to compute `sens()`.
#> Caused by error:
#> ! `truth` should be a factor, not a `numeric`.
```

<sup>Created on 2023-05-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/fair-aaa.R
```diff
@@ -0,0 +1,137 @@
...
+  diff(range(x$.estimate))
+}
+
+max_positive_rate_diff <- function(x, ...) {
+  metric_values <- vctrs::vec_split(x, x$.metric)
+
+  positive_rate_diff <- vapply(metric_values$val, diff_range, numeric(1), ...)
+
+  max(positive_rate_diff)
+}
```
EmilHvitfeldt, line 137: Unless you are expecting these function to be used across many new metrics, I would move them closer to where there are used.

R/fair-demographic_parity.R
```diff
@@ -0,0 +1,47 @@
+#' Demographic parity
```
EmilHvitfeldt, line 1: In this and the other metrics, I like what you are doing with the templating. I still think it would be good to make sure we include a one-liner describing what the function does. e.i.

> demographic parity is calculated as the difference between the largest and smallest value of detection_prevalence that were calculated for each group

R/fair-aaa.R
```diff
@@ -0,0 +1,137 @@
...
+        res
+      }
+    res <- new_class_metric(res, direction = "minimize")
+    attr(res, "by") <- by_str
+    res
+  }
+}
+
+diff_range <- function(x, ...) {
+  diff(range(x$.estimate))
```
EmilHvitfeldt, line 128: this feels like a major nitpick, but we can calculate this with `min()` and `max()`

``` r
f1 <- function(x) {
  diff(range(x))
}

f2 <- function(x) {
  max(x) - min(x)
}

x <- rnorm(10)

bench::mark(
  f1(x),
  f2(x)
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 f1(x)        2.67µs   3.03µs   238952.        0B     23.9
#> 2 f2(x)      327.83ns 369.04ns  1965169.        0B    197.
```

_pkgdown.yml
```diff
@@ -68,6 +68,13 @@ reference:
     - iic
     - poisson_log_loss
 
+  - title: Fairness Metrics
+    contents:
+    - fairness_metric
```
EmilHvitfeldt, line 73: this function should be listed under `Development Functions`

EmilHvitfeldt (commented):
Super excited about these changed!

_pkgdown.yml
juliasilge: It's significantly higher level than the functions for creating new metrics, and I think we suspect more likely to be used by "regular" users.

R/aaa-metrics.R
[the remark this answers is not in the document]
simonpcouch: From the commit [b98b2ac](https://github.com/tidymodels/yardstick/pull/434/commits/b98b2ac94f4a2458445462b279af8b5d1792d45e):

> The forthcoming metrics execute these functions via their constructor and thus introduce errors at `load_all()`. We could also instead rename the fairness files to be run last, e.g. `zzz-fair.R`.

R/fair-aaa.R
```diff
@@ -0,0 +1,137 @@
...
+#'     .fn = detection_prevalence,
+#'     .name = "demographic_parity_ratio",
+#'     .post = ratio_range
+#'   )
+#'
+#' @export
+fairness_metric <- function(.fn, .name, .post) {
+  if (rlang::is_missing(.fn) || !inherits_any(.fn, c("metric", "metric_set"))) {
+    rlang::abort("`.fn` must be a metric function or metric set.")
+  }
```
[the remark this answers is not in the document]
simonpcouch: Ah, nice! Thanks.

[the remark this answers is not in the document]
simonpcouch: [6998108](https://github.com/tidymodels/yardstick/pull/434/commits/6998108e7b52f34348230981041466b1abd0f005) :)

R/fair-aaa.R
simonpcouch: That's a big ol difference. :) You got it! [a9fc2e0](https://github.com/tidymodels/yardstick/pull/434/commits/a9fc2e08f62fb91e4523a15c15f6a9cf08aae95e)

R/fair-aaa.R
simonpcouch: `diff_range()` is just used in two places, though `max_positive_rate_diff()` just one. I wasn't sure whether we would eventually want to supply a larger suite of fairness metrics than these three—if we feel good about just supplying these canonical ones, `max_positive_rate_diff()` should definitely move to `equalized_odds()`!

R/fair-aaa.R
simonpcouch: Transitioned from `summarize()` in [27f53e9](https://github.com/tidymodels/yardstick/pull/434/commits/27f53e9fed2ac1627702b267bd6d6177cfdb5653)! Thought that error message still looks the same.

simonpcouch: Errors from `demographic_parity()` and `equal_opportunity()` look a bit different since their `.fn` is a metric rather than a metric set. With those:

``` r
library(yardstick)

data(hpc_cv)

equal_opportunity_resample <- equal_opportunity(Resample)

hpc_cv |>
  equal_opportunity_resample(truth = VF, estimate = pred)
```

It'd be nice if that could say `equal_opportunity_resample()` instead of `.fn()`, since the user maybe need not know about the existence of `fairness_metric()`, but this might require some cleverness in passing the `call` to metric functions like `sens()` or `spec()`.

R/fair-aaa.R
```diff
@@ -0,0 +1,137 @@
...
+
+        if (length(gp_vars) > 0) {
+          res <- dplyr::group_by(res, !!!dplyr::groups(data), .add = FALSE)
+        }
+
+        res <-
+          dplyr::summarize(
+            res,
+            .metric = .name,
+            !!".by" := by_str,
```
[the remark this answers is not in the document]
simonpcouch: Resolved via [27f53e9](https://github.com/tidymodels/yardstick/pull/434/commits/27f53e9fed2ac1627702b267bd6d6177cfdb5653).

R/fair-aaa.R
EmilHvitfeldt: I fully support trying to apply cleverness! 😄

R/fair-aaa.R
```diff
@@ -0,0 +1,181 @@
...
+    abort("`.name` must be a string.")
+  }
+  if (is_missing(.post) || !is_function(.post)) {
+    abort("`.post` must be a function.")
+  }
+
+  function(by) {
+    by_str <- as_string(enexpr(by))
+    res <-
+      function(data, ...) {
```
simonpcouch, line 119: The `...` in this argument are passed to both `.fn` and to `.post`. We could:

* Remove the `...` argument to `.post`, since defining a problem-context-aware function means that all of the information you need likely lives in the metric set or can be hard-coded into `.post`
* Resolve `formals(.fn)` as the formals of this function, which would fully address the case where `.fn` is one yardstick function like `sens` or `detection_prevalence`; those functions don't use their dots, and `.post` would only get the dots it needs (i.e. doesn't need to know what to do with `truth` or `estimate`)
* Add a check a few lines above here to ensure that `.post` has a `...` in that argument. In that case, `truth`/`estimate`/etc are not `force()`d and users shouldn't run into issues as long as they don't define arguments to `.post` with the same name as an argument to `.fn`.

R/fair-aaa.R
```diff
@@ -0,0 +1,181 @@
...
+#' function except that it will look for a `"gender"` column in
+#' the data it's supplied.
+#'
+#' In addition to the examples below, see the documentation on the
+#' return value of fairness metrics like [demographic_parity()],
+#' [equal_opportunity()], or [equalized_odds()] to learn more about how the
+#' output of this function can be used.
+#'
+#' @examples
+#' data(hpc_cv)
```
simonpcouch, line 73: If we decide to move forward with a fairness-oriented name, it'd be great if we could use some example data here that has a plausibly "sensitive" attribute. yardstick doesn't `Suggests` modeldata at the moment, which has some options. `infer::gss` would also work well here.

R/fair-aaa.R
simonpcouch: See https://github.com/tidymodels/yardstick/actions/runs/4996847332/jobs/8950515558#step:6:178 for an example of how this comes up.

R/fair-aaa.R
```diff
@@ -0,0 +1,181 @@
...
+  }
+  if (is_missing(.post) || !is_function(.post)) {
+    abort("`.post` must be a function.")
+  }
+
+  function(by) {
+    by_str <- as_string(enexpr(by))
+    res <-
+      function(data, ...) {
+        gp_vars <- dplyr::group_vars(data)
```
simonpcouch, line 120: We should check that `by_str` isn't in `gp_vars` here.

R/fair-aaa.R
```diff
@@ -0,0 +1,137 @@
...
+#'
+#' demographic_parity_ratio <-
+#'   fairness_metric(
+#'     .fn = detection_prevalence,
+#'     .name = "demographic_parity_ratio",
+#'     .post = ratio_range
+#'   )
+#'
+#' @export
+fairness_metric <- function(.fn, .name, .post) {
```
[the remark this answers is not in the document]
simonpcouch: Notes from the group meeting:
* Max mentioned that it might be nice to prefix whatever this function name is with `create_` or some other eliciting verb to indicate that this is a function factory, and others agreed
* I suggested `disparity_metric` as a descriptor for this type of metric that doesn't have as strong of a social connotation—seems like "disparity" could describe differences across groups regardless of whether that group is regarded as a sensitive feature

R/fair-aaa.R
simonpcouch: Addressed in [f1d663f](https://github.com/tidymodels/yardstick/pull/434/commits/f1d663fdcbd0d3ab7b087390d710711a25c29e19)!

R/fair-aaa.R
```diff
@@ -0,0 +1,181 @@
...
+    res <- new_class_metric(res, direction = "minimize")
+    attr(res, "by") <- by_str
+    res
+  }
+}
+
+diff_range <- function(x, ...) {
+  estimates <- x$.estimate
+
+  max(estimates) - min(estimates)
```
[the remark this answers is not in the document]
simonpcouch: Possibly? Should test interactions with `na_rm` here.

R/fair-aaa.R
simonpcouch: Addressed in [45f57ff](https://github.com/tidymodels/yardstick/pull/434/commits/45f57ff6a1480a53d66ed1bbfe8cbba4d7b5762d) via the first bulleted option.

R/fair-aaa.R
```diff
@@ -0,0 +1,181 @@
...
+          out[[i]] <- tibble::new_tibble(elt_out)
+        }
+
+        group_keys <- vctrs::vec_rep_each(group_keys, times = list_sizes(out))
+        out <- vec_rbind(!!!out)
+        out <- vec_cbind(group_keys, out)
+
+        out
+      }
+    res <- new_class_metric(res, direction = "minimize")
```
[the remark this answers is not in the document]
simonpcouch: Addressed in [6fad6a4](https://github.com/tidymodels/yardstick/pull/434/commits/6fad6a46d6bdfb015177e8091e52ef031c3f8713).

R/fair-aaa.R
simonpcouch: Addressed in [82717fa](https://github.com/tidymodels/yardstick/pull/434/commits/82717fadf5409b9630ed02c4ba95aa1e8e40fcee). :)

R/fair-demographic_parity.R
simonpcouch: Addressed in [a2dfe36](https://github.com/tidymodels/yardstick/pull/434/commits/a2dfe363a9fed26c5f204f9c87237d03f957df0b)!

R/fair-aaa.R
simonpcouch: Addressed in [375269b](https://github.com/tidymodels/yardstick/pull/434/commits/375269bdf88f6e494f4506cb282d2e292dfef40f)! Thanks for the nudge on this. :)

``` r
library(yardstick)

data(hpc_cv)

equalized_odds_resample <- equalized_odds(Resample)

hpc_cv |>
  equalized_odds_resample(truth = VF, estimate = pred)
#> Error in `equalized_odds()`:
#> ! `truth` should be a factor, not a `numeric`.
```

<sup>Created on 2023-06-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

juliasilge:
FWIW I also think it is a good thing (not a problem) that the syntax emphasizes that `demographic_parity(cyl)` is the same kind of thing (a function) as `roc_auc`. Using a subclass would be a good option to help with how folks handle these fairness metrics, and relates back to what I mentioned before about a `print` method for metrics in general.

R/fair-aaa.R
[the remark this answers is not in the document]
juliasilge: I like the idea of using `create_*` here, but it is a departure compared to the other function factories in yardstick (of which there are a lot, like `metric_tweak()`, `metric_set()`, and so forth). Do you think it's better to stay more similar to the naming conventions of yardstick, or to use something like `create_*`?

I have a mild preference for something like `create_groupwise_metric()` because I think there is more ML community vocabulary around what "groupwise" means. The word "disparity" makes me think about the specific metric [disparate impact](https://www.ibm.com/docs/en/cloud-paks/cp-data/4.5.x?topic=monitor-fairness-metric-disparate-impact). That being said, my opinion is not super strong here.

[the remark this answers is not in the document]
simonpcouch: Good points--if we want to look to other function factories in the package, maybe the parallel we might want to draw is with `new_metric()`? Something like `new_groupwise_metric()`?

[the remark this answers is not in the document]
juliasilge: I like that a lot, `new_groupwise_metric()` 👍

EmilHvitfeldt:
I think part of the confusion arrises on how we use these functions.

If we change

```r
metric_set(
  roc_auc, 
  accuracy, 
  demographic_parity(cyl)
)
```

To the following, then we are no longer using `()` inside `metric_set()`

```r
demographic_parrity_cyl <- demographic_parity(cyl)

metric_set(
  roc_auc, 
  accuracy, 
  demographic_parrity_cyl
)
```

In the end this code is a little tricky because we are using a function factory (`demographic_parity()`) to create a function that is passed to a different function factory (`metric_set()`).

I'm happy with the interface as it stands right now. And I think we can get a lot of leeway by using better printing, plus we can add some robust error handling in `metric_set()`

simonpcouch:
`metric_set()` will now error (extra) informatively when passed an unevaluated fairness metric function! Interactively, "group-wise metric" is a hyperlink to `?yardstick::fairness_metric()`.

``` r
library(yardstick)
  
metric_set(demographic_parity)
#> Error in `metric_set()`:
#> ! The input `demographic_parity` is a group-wise metric
#>   (`?yardstick::fairness_metric()`) factory and must be passed a data-column
#>   before addition to a metric set.
#> ℹ Did you mean to type e.g. `demographic_parity(col_name)`?

metric_set(demographic_parity, equal_opportunity)
#> Error in `metric_set()`:
#> ! The inputs `demographic_parity` and `equal_opportunity` are group-wise
#>   metric (`?yardstick::fairness_metric()`) factories and must be passed a
#>   data-column before addition to a metric set.
#> ℹ Did you mean to type e.g. `demographic_parity(col_name)`?
```

<sup>Created on 2023-06-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/fair-aaa.R
```diff
@@ -0,0 +1,220 @@
...
+#'
+#' demographic_parity_ratio <-
+#'   new_groupwise_metric(
+#'     .fn = detection_prevalence,
+#'     .name = "demographic_parity_ratio",
+#'     .post = ratio_range
+#'   )
+#'
+#' @export
+new_groupwise_metric <- function(.fn, .name, .post, direction = "minimize") {
```
simonpcouch, line 113: The biggest remaining question for me is the naming of these arguments. Some stray thoughts:

* `.fn` was introduced for compatibility with `metric_tweak()`. Now that this function is named more similarly to `new_metric()`, and has a `direction` argument, maybe this should be `fn`?
* Davis suggested that `.post` could be `aggregate` or `.aggregate`. I do think "aggregate" or "summarize" is a more natural word for what's going on here, and at the same time am a little hesitant to name this argument in that way because they are functions and the tidy-style-aligned way to write about them would be `aggregate()` and `summarize()`, which suggest the actual stats and dplyr verbs to me.
* `.name` can switch to `name` if other arguments lose their `.` prefix.

simonpcouch:
@EmilHvitfeldt Some big-picture for re-review:

I'd like to discuss + make some changes related to https://github.com/tidymodels/yardstick/pull/434#discussion_r1242171885 before merging!

The next steps for this work are 1) a vignette (/article?) in yardstick using a proper example to demonstrate using fairness metrics, 2) a tidymodels.org blog post with a separate example that uses these metrics while tuning and makes use of custom fairness metrics, and then 3) a tidyverse blog post giving the bigger picture of this work and linking out to 1) and 2). If you'd like 1) to happen in this PR, that totally works, otherwise I'll plan to follow up on this PR with a smaller, separate one. 

After 1-3, it's time to delve into functionality for mitigation.🏄

R/fair-aaa.R
EmilHvitfeldt: 1. I think `fn` fits better, it aligns with `new_metric()` and the `*_metric_summarizer()` functions. Which makes sense, since this function is used to create metrics, not modify existing metrics which is what `metric_tweak()` is doing.
2. I would go with `aggregate`. We already use `summarize` in `collect_metrics()` and I think it would be confusing if it would be used differently here. `post`/`.post` might be more useful for post processing in workflows terminology.
3. Yes, switch to `name`. it will also align with `new_metric()` and the `*_metric_summarizer()` functions.

EmilHvitfeldt:
I would like the vignette work to happen in a new PR. It will be cleaner that way!

R/fair-aaa.R
```diff
@@ -0,0 +1,221 @@
...
+#' grouping by the variable passed to `by` and then summarizing the per-group
+#' metric estimates across groups using the function passed as the
+#' `aggregrate` argument.
+#'
+#' @param fn A yardstick metric function or metric set.
+#' @param name The name of the metric to place in the `.metric` column
+#' of the output.
+#' @param aggregrate A function to summarize the generated metric set results
+#' `x`. In many cases, `~diff(range(x$.estimate))` or
+#' `~r <- range(x$.estimate); r[1]/r[2]`.
```
EmilHvitfeldt, line 22: This documentation should be expanded. I would look at `?purrr::map()` for inspiration.

Right now, the examples listed are technically not functions

simonpcouch: Thanks for bringing this up—addressed in [4436785](https://github.com/tidymodels/yardstick/pull/434/commits/4436785759021d88db4360ce7f6f62b2d419f69f). Opted to describe the input and output type more thoroughly and refer to other sections for worked examples.

simonpcouch:
Realizing that `aggregate` is misspelled `aggregrate`---one moment.🙃

EmilHvitfeldt:
Pkgdown CI doesn't work because http://r-project.org/ is down. I'm merging anyways

PR #453, opened 2023-10-29
add vignette on group-aware and -wise behavior

simonpcouch:
Related to #434.

### DIFF
--- a/vignettes/grouping.Rmd
+++ b/vignettes/grouping.Rmd
@@ -1,0 +1,172 @@
+---
+title: "Grouping behavior in yardstick"
+author: "Simon Couch"
+date: "`r Sys.Date()`"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Grouping behavior in yardstick}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r setup, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+The 1.3.0 release of yardstick introduced an implementation for _group-wise metrics_. The use case motivating the implementation of this functionality is _fairness metrics_, though group-wise metrics have applications beyond that domain. Fairness metrics quantify the degree of disparity in a metric value across groups. To learn more about carrying out fairness-oriented analyses with tidymodels, see the blog post on the tidymodels website. This vignette will instead focus on group-wise metrics generally, clarifying the meaning of "group-wise" and demonstrating functionality with an example dataset.
+
+<!-- TODO: link to forthcoming tidymodels blog post -->
+
+```{r pkgs, message = FALSE}
+library(yardstick)
+library(dplyr)
+
+data("hpc_cv")
+```
+
+# Group-awareness
+
+Even before the implementation of group-wise metrics, _all_ yardstick metrics had been _group-aware_. By group-aware, we mean that when passed grouped data, a metric will return metric values calculated for each group.
+
+To demonstrate, we'll make use of the `hpc_cv` data set, containing class probabilities and class predictions for a linear discriminant analysis fit to the HPC data set of Kuhn and Johnson (2013). The model is evaluated against a 10 fold cross-validation, and the predictions for all folds are included. 
+
+```{r hpc-cv}
+tibble(hpc_cv)
+```
+
+For the purposes of this vignette, we'll also add a column `batch` to the data and select off the columns for the class probabilities, which we don't need.
+
+```{r hpc-modify}
+set.seed(1)
+
+hpc <-
+  tibble(hpc_cv) %>%
+  mutate(batch = sample(c("a", "b"), nrow(.), replace = TRUE)) %>%
+  select(-c(VF, F, M, L))
+
+hpc
+```
+
+If we wanted to compute the accuracy of the first resampled model, we could write:
+
+```{r acc-1}
+hpc %>% 
+  filter(Resample == "Fold01") %>%
+  accuracy(obs, pred)
+```
+
+The metric function returns one row, giving the `.metric`, `.estimator`, and `.estimate` for the whole data set it is passed.
+
+If we instead group the data by fold, metric functions like `accuracy` will know to compute values for each group; in the input, each row will correspond to a Resample.
+
+```{r hpc-cv-2}
+hpc %>% 
+  group_by(Resample) %>%
+  accuracy(obs, pred)
+```
+
+Note that the first row, corresponding to `Resample`, gives the same value as manually filtering for the observations corresponding to the first resample and then computing the accuracy.
+
+This behavior is what we mean by group-awareness. When passed grouped data, metric functions will return values for each group.
+
+# Group-wise metrics
+
+Group-wise metrics are associated with a data-column such that, when passed data with that column, the metric will temporarily additionally group by that column, compute values for each of the groups defined by the column, and then aggregate the values computed for the temporary grouping back to the level of the inputted data's grouping.
+
+More concretely, let's turn to an example where there is no pre-existing grouping in the data. Consider the portion of the HPC data pertaining to the first resample:
+
+```{r res-1}
+hpc %>% 
+  filter(Resample == "Fold01")
+```
+
+Suppose that the `batch`es in the data represent two groups for which model performance ought not to differ. To quantify the degree to which model performance differs for these two groups, we could compute accuracy values for either group separately, and then take their difference. First, computing accuracies:
+
+```{r acc-by-group}
+acc_by_group <- 
+  hpc %>% 
+  filter(Resample == "Fold01") %>%
+  group_by(batch) %>%
+  accuracy(obs, pred)
+
+acc_by_group
+```
+
+Now, taking the difference:
+
+```{r diff-acc}
+diff(c(acc_by_group$.estimate[2], acc_by_group$.estimate[1]))
+```
+
+Group-wise metrics encode the `group_by()` and subtraction step shown above into a yardstick metric. We can define a new group-wise metric with the `new_groupwise_metric()` function:
+
+```{r}
+accuracy_diff <-
+  new_groupwise_metric(
+    fn = accuracy,
+    name = "accuracy_diff",
+    aggregate = function(acc_by_group) {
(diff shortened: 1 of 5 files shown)


vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
+---
+title: "Grouping behavior in yardstick"
```
simonpcouch, line 2: I had originally thought maybe this vignette would be called "Group-wise metrics" but it ended up being a bit more general than that.

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+      diff(c(acc_by_group$.estimate[2], acc_by_group$.estimate[1]))
+    }
+  )
+```
+
+* The `fn` argument is the yardstick metric that will be computed for each data group.
+* The `name` argument gives the name of the new metric we've created; we'll call ours "accuracy difference."
+* The `aggregate` argument is a function defining how to go from `fn` output by group to a single numeric.
+
+The output, `accuracy_diff`, is a function subclass called a `metric_factory`:
```
simonpcouch, line 121: Trying to talk about functions as objects while adhering to tidyverse style is a bit difficult here. I'm game for however yall want to approach this. :)

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+```{r pkgs, message = FALSE}
+library(yardstick)
+library(dplyr)
+
+data("hpc_cv")
+```
+
+# Group-awareness
+
+Even before the implementation of group-wise metrics, _all_ yardstick metrics had been _group-aware_. By group-aware, we mean that when passed grouped data, a metric will return metric values calculated for each group.
```
EmilHvitfeldt, line 32: I understand what it meant with "when passed grouped data", but I think it takes a couple of reads to understand.

What do you think about the following rephrase? 

```suggestion
Even before the implementation of group-wise metrics, _all_ yardstick metrics had been _group-aware_. When grouped data is passed to a group-aware metric, it willwill return metric values calculated for each group.
```

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+To demonstrate, we'll make use of the `hpc_cv` data set, containing class probabilities and class predictions for a linear discriminant analysis fit to the HPC data set of Kuhn and Johnson (2013). The model is evaluated against a 10 fold cross-validation, and the predictions for all folds are included.
```
EmilHvitfeldt, line 34:
```suggestion
To demonstrate, we'll make use of the `hpc_cv` data set, containing class probabilities and class predictions for a linear discriminant analysis fit to the HPC data set of Kuhn and Johnson (2013). The model is evaluated against a 10-fold cross-validation, and the predictions for all folds are included. 
```

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+```{r hpc-cv-2}
+hpc %>% 
+  group_by(Resample) %>%
+  accuracy(obs, pred)
+```
+
+Note that the first row, corresponding to `Resample`, gives the same value as manually filtering for the observations corresponding to the first resample and then computing the accuracy.
+
+This behavior is what we mean by group-awareness. When passed grouped data, metric functions will return values for each group.
```
EmilHvitfeldt, line 73: Similar vibe to earlier comment

```suggestion
This behavior is what we mean by group-awareness. When grouped data is passed to group-aware metric functions they will return values for each group.
```

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+```{r metric-classes}
+class(accuracy)
+
+class(accuracy_diff_by_batch)
+```
+
+<!-- TODO: once a print method is added, we can print this out and the meaning of "this is just a yardstick metric" will be clearer -->
+
+We can use the `accuracy_diff_by_batch()` metric in the same way that we would use `accuracy()`. On it's own:
```
EmilHvitfeldt, line 143:
```suggestion
We can use the `accuracy_diff_by_batch()` metric in the same way that we would use `accuracy()`. On its own:
```

EmilHvitfeldt (approved):
Few comments!

vignettes/grouping.Rmd
EmilHvitfeldt: with a title like that, we might wanna preface a little earlier that this isn't about grouped data sets

simonpcouch: with a title like the one currently suggested or the previous "Group-wise metrics"? and in some ways, this is indeed about grouped data.

simonpcouch:
Much appreciated! Accepted all suggestions.

vignettes/grouping.Rmd
EmilHvitfeldt: i get worried, and confused myself a little bit, when we have this new methods, that share close names with group_by() related stuff. 

I was thinking of having a "when we say X we talk about ABC, and when we say Y we talk about XYZ"

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+library(dplyr)
+
+data("hpc_cv")
+```
+
+# Group-awareness
+
+Even before the implementation of group-wise metrics, _all_ yardstick metrics had been _group-aware_. When grouped data is passed to a group-aware metric, it will return metric values calculated for each group.
+
+To demonstrate, we'll make use of the `hpc_cv` data set, containing class probabilities and class predictions for a linear discriminant analysis fit to the HPC data set of Kuhn and Johnson (2013). The model is evaluated against a 10-fold cross-validation, and the predictions for all folds are included.
```
juliasilge, line 34:
```suggestion
To demonstrate, we'll make use of the `hpc_cv` data set, containing class probabilities and class predictions for a linear discriminant analysis fit to the HPC data set of Kuhn and Johnson (2013). The model is evaluated via 10-fold cross-validation, and the predictions for all folds are included. 
```

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+```{r acc-1}
+hpc %>% 
+  filter(Resample == "Fold01") %>%
+  accuracy(obs, pred)
+```
+
+The metric function returns one row, giving the `.metric`, `.estimator`, and `.estimate` for the whole data set it is passed.
+
+If we instead group the data by fold, metric functions like `accuracy` will know to compute values for each group; in the input, each row will correspond to a Resample.
```
juliasilge, line 63:
```suggestion
If we instead group the data by fold, metric functions like `accuracy` will know to compute values for each group; in the output, each row will correspond to a Resample.
```

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+```{r hpc-cv-2}
+hpc %>% 
+  group_by(Resample) %>%
+  accuracy(obs, pred)
+```
+
+Note that the first row, corresponding to `Resample`, gives the same value as manually filtering for the observations corresponding to the first resample and then computing the accuracy.
```
juliasilge, line 71:
```suggestion
Note that the first row, corresponding to `Fold01 `, gives the same value as manually filtering for the observations corresponding to the first resample and then computing the accuracy.
```

Is that what is intended here?

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+This behavior is what we mean by group-awareness. When grouped data is passed to group-aware metric functions, they will return values for each group.
+
+# Group-wise metrics
+
+Group-wise metrics are associated with a data-column such that, when passed data with that column, the metric will temporarily additionally group by that column, compute values for each of the groups defined by the column, and then aggregate the values computed for the temporary grouping back to the level of the inputted data's grouping.
```
juliasilge, line 77:
```suggestion
Group-wise metrics are associated with a data-column such that, when passed data with that column, the metric will temporarily group by that column, compute values for each of the groups defined by the column, and then aggregate the values computed for the temporary grouping back to the level of the input data's grouping.
```

You address being group aware at the end very clearly, and I think "additionally" is kind of confusing here.

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+acc_by_group
+```
+
+Now, taking the difference:
+
+```{r diff-acc}
+diff(c(acc_by_group$.estimate[2], acc_by_group$.estimate[1]))
+```
+
+Group-wise metrics encode the `group_by()` and subtraction step shown above into a yardstick metric. We can define a new group-wise metric with the `new_groupwise_metric()` function:
```
juliasilge, line 104:
```suggestion
Group-wise metrics encode the `group_by()` and aggregation step (in this case, subtraction) shown above into a yardstick metric. We can define a new group-wise metric with the `new_groupwise_metric()` function:
```

vignettes/grouping.Rmd
```diff
@@ -0,0 +1,172 @@
...
+    name = "accuracy_diff",
+    aggregate = function(acc_by_group) {
+      diff(c(acc_by_group$.estimate[2], acc_by_group$.estimate[1]))
+    }
+  )
+```
+
+* The `fn` argument is the yardstick metric that will be computed for each data group.
+* The `name` argument gives the name of the new metric we've created; we'll call ours "accuracy difference."
+* The `aggregate` argument is a function defining how to go from `fn` output by group to a single numeric.
```
juliasilge, line 119:
```suggestion
* The `aggregate` argument is a function defining how to go from `fn` output by group to a single numeric value.
```

juliasilge (approved):
Looking fantastic! 🎉 

This is _such_ a tiny nit, but I prefer "groupwise" to "group-wise":
- https://www.merriam-webster.com/dictionary/groupwise
- https://www.wordnik.com/words/groupwise

simonpcouch:
On board for all of your changes, thank you @juliasilge!

Ah, no, that's super important to get right! I'm all for transitioning from "group-wise" to "groupwise". Any thoughts on whether we should then write "group aware" instead of "group-aware"?

juliasilge:
Seems like "group-aware" is more common, [like this](https://proceedings.mlr.press/v202/hosseini23a.html).

simonpcouch:
Awesome, we'll keep that, then.

After putting [9145ba1](https://github.com/tidymodels/yardstick/pull/453/commits/9145ba1231c5bfddc9c058f5f5f3ba86ee173fdc) together, I do also appreciate that the way we write "groupwise" in text aligns with how it appears in code. :)

Thanks again.🐙

PR #518, opened 2024-10-24
Enable checks without suggested packages

topepo:
Emulates the "no suggests" testing variant by CRAN.


EmilHvitfeldt (approved):
amazing! sooo clean 🧼

PR #520, opened 2024-10-25
All snapshots

EmilHvitfeldt:
All errors are now snapshotted whenever possible, otherwise they are labeled as unreachable

### DIFF
--- a/tests/testthat/test-template.R
+++ b/tests/testthat/test-template.R
@@ -14,7 +14,7 @@ test_that("missing values in case weights are considered by `na_rm`", {
   )
 })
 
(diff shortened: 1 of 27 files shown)


tests/testthat/test-template.R
```diff
@@ -14,6 +14,17 @@ test_that("missing values in case weights are considered by `na_rm`", {
   )
 })
 
+test_that("errors are thrown truth or estimate selects more than 1 column", {
```
hfrick, line 17:
```suggestion
test_that("errors are thrown if truth or estimate selects more than 1 column", {
```

hfrick (approved):
Looks good! 🧹✨

Issue #521, opened 2024-10-26
pass `call` to `weighted_table()`


EmilHvitfeldt:
wait for https://github.com/tidymodels/hardhat/issues/273

PR #522, opened 2024-10-26
pass calls around

EmilHvitfeldt:
This PR passes the calls around in all missing cases, expect for `weighted_table()` which will be tracked with https://github.com/tidymodels/yardstick/issues/521

### DIFF
--- a/R/metric-tweak.R
+++ b/R/metric-tweak.R
@@ -85,7 +85,7 @@ metric_tweak <- function(.name, .fn, ...) {
 
 # ------------------------------------------------------------------------------
 
-check_protected_names <- function(fixed) {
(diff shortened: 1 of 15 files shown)


R/metric-tweak.R
```diff
@@ -94,7 +94,8 @@ check_protected_names <- function(fixed) {
   }
 
   cli::cli_abort(
-    "Arguments passed through {.arg ...} cannot be named any of: {protected}."
+    "Arguments passed through {.arg ...} cannot be named any of: {protected}.",
```
topepo, line 97:
```suggestion
    "Arguments passed through {.arg ...} cannot be named any of: {.arg {protected}}.",
```

PR #555, opened 2025-09-19
Interactively asks user to install `probably` if it is missing and required

JavOrraca:
Additionally, fixed some typos from "not a a " to "not a "


EmilHvitfeldt:
Thank you!

PR #567, opened 2025-12-15
Royston metric

hfrick:
Follow-up to #564, now adding a new metric of the new `linear_pred_surival_metric` type

This PR add the metric from Royston and Sauerbrei (2004) and extends it to work with case weights. Two comments on statistics aspects below, please weigh in on whether or not this seems sensible to you.

For the basic implementation without case weights, I compare it to the one from the `survival` package. In `survival::royston()`, the predictions get rescaled:
https://github.com/therneau/survival/blob/20dd980acc7b051b1b4a6ef5988b8f130c81474b/R/royston.R#L24-L29 I'm not sure why though, I didn't see that part in the Royston and Sauerbrei paper. Given that we don't even have the model formula at this point, I've gone ahead with the predictions as they are given to the yardstick function.

For case weights: the metric function from survival does not take case weights, so this here is my own take on it. Generally, we pass case weights to the metric calculation from the tuning process if they are frequency weights (important weights are for model fitting only). So I've treated case weights here as frequency weights. 
The metric is a two-parter: calculate Blom's approximation to rankits (based on ranks), then fit a Cox model on them and use the coefficient to calculate the metric. Both of those elements should be affected by frequency weights imo, so that's how I've implemented it. While fitting the Cox model could take importance weights, I don't think they'd make much sense for the rankits, so that assumption of _frequency_ weights is built into `normal_score_blom()`.

### DIFF
--- a/R/surv-royston.R
+++ b/R/surv-royston.R
@@ -109,7 +109,10 @@ royston_survival_impl <- function(truth, estimate, case_weights) {
 }
 
 normal_score_blom <- function(x, case_weights) {
-  tibble::tibble(
+  # includes observations with zero weights
+  x_0 <- tibble::tibble(.row = seq_along(x), x = x)
+
+  rankits <- tibble::tibble(
     .row = rep(seq_along(x), times = case_weights),
     x = rep(x, times = case_weights),
   ) |>
@@ -122,6 +125,8 @@ normal_score_blom <- function(x, case_weights) {
     ) |>
     # average over ties
     dplyr::mutate(s = mean(.data$z), .by = "x") |>
-    dplyr::slice(1, .by = .row) |>
+    dplyr::slice(1, .by = .row)
+
+  dplyr::left_join(x_0, rankits, by = c(".row")) |>
     dplyr::pull("s")
 }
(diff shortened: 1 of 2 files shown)


R/surv-royston.R
```diff
@@ -0,0 +1,124 @@
...
+  tibble::tibble(
+    .row = rep(seq_along(x), times = case_weights),
+    x = rep(x, times = case_weights),
+  ) |>
+    dplyr::mutate(
+      x_first = rank(.data$x, ties.method = "first"),
+      # does not need kappa (from Royston & Sauerbrei 2004) because it'll
+      # "disappear" into the baseline hazard of the Cox model in
+      # `royston_survival_impl()`
+      z = stats::qnorm((.data$x_first - 3 / 8) / (dplyr::n() + 0.25))
```
EmilHvitfeldt: I know they are the same but this makes it look even more like the formula in the paper

```suggestion
      z = stats::qnorm((.data$x_first - 3 / 8) / (dplyr::n() + 1 / 4))
```

EmilHvitfeldt (approved):
It is all implemented very nicely! following the existing frameworks.

hfrick:
Just chatted with Max and his 👍 on the PR description was the 👍 on the case weights handling so I'm gonna merge this now!

Issue #504, opened 2024-04-28
`metric_set()` has a less useful error message than classification metric functions when `truth` or `estimate` is not a factor


mikemahoney218:
## The problem

For classification metrics, when `truth` or `estimate` is not a factor, `metric_set()` gives a much less informative error than the classification metric itself. 

## Reproducible example

``` r
yardstick::precision(Orange, age, Tree)
#> Error in `yardstick::precision()`:
#> ! `truth` should be a factor, not a a double vector.
#> Backtrace:
#>      ▆
#>   1. ├─yardstick::precision(Orange, age, Tree)
#>   2. └─yardstick:::precision.data.frame(Orange, age, Tree)
#>   3.   └─yardstick::class_metric_summarizer(...)
#>   4.     ├─rlang::inject(...)
#>   5.     ├─base::withCallingHandlers(...)
#>   6.     └─yardstick (local) fn(...)
#>   7.       └─yardstick::check_class_metric(...)
#>   8.         └─yardstick:::validate_factor_truth_factor_estimate(...)
#>   9.           └─cli::cli_abort(...)
#>  10.             └─rlang::abort(...)
yardstick::metric_set(yardstick::precision)(Orange, age, Tree)
#> Error in `metric_set()`:
#> ! Failed to compute `precision()`.
#> Caused by error:
#> ! Must select at least one item.
#> Backtrace:
#>      ▆
#>   1. ├─(yardstick::metric_set(yardstick::precision))(Orange, age, Tree)
#>   2. │ └─base::mapply(...)
#>   3. │   └─yardstick (local) `<fn>`(dots[[1L]][[1L]], dots[[2L]][[1L]])
#>   4. │     ├─base::tryCatch(...)
#>   5. │     │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>   6. │     │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>   7. │     │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
#>   8. │     └─rlang::eval_tidy(expr, data = data, env = env)
#>   9. ├─yardstick (local) `<clss_mtr>`(...)
#>  10. └─yardstick:::precision.data.frame(...)
#>  11.   └─yardstick::class_metric_summarizer(...)
#>  12.     └─yardstick:::yardstick_eval_select(...)
#>  13.       └─tidyselect::eval_select(...)
#>  14.         └─tidyselect:::eval_select_impl(...)
[...]

PR #603, opened 2026-01-21
Estimate arg named metric set

EmilHvitfeldt:
to close https://github.com/tidymodels/yardstick/issues/504

``` r
yardstick::precision(Orange, age, Tree)
#> Error in `yardstick::precision()`:
#> ! `truth` should be a factor, not a a double vector.

yardstick::metric_set(yardstick::precision)(Orange, age, Tree)
#> Error:
#> ! `estimate` is required for class metrics but was not provided.
#> ℹ The `estimate` argument must be named because it comes after `...`.
#> ℹ Example: `my_metrics(data, truth, estimate = my_column)`

my_set <- yardstick::metric_set(yardstick::precision)

my_set(Orange, age, Tree)
#> Error in `my_set()`:
#> ! `estimate` is required for class metrics but was not provided.
#> ℹ The `estimate` argument must be named because it comes after `...`.
#> ℹ Example: `my_metrics(data, truth, estimate = my_column)`
```

<sup>Created on 2026-01-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/aaa-metric_set.R
+++ b/R/aaa-metric_set.R
@@ -315,8 +315,7 @@ make_prob_class_metric_function <- function(fns) {
         c(
           "!" = "{.arg estimate} is required for class metrics but was not 
                  provided.",
-          "i" = "The {.arg estimate} argument must be named because
-                 it comes after {.arg ...}.",
+          "i" = "In a metric set, the {.arg estimate} argument must be named.",
           "i" = "Example: {.code my_metrics(data, truth, estimate = my_column)}"
         ),
         call = current_env()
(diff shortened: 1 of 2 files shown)


hfrick (approved):
This is much better than before! :raised_hands: 

One suggestion: given that the `estimate` argument only has to be named when it's in a metric set (but not when using the metric directly), I would prioritize squeezing that into the error message. The why is nice to have and I only took it out to make the error about the same length as before (assuming that the longer it gets, the less likely it gets read).

```
my_set(Orange, age, Tree)
#> Error in `my_set()`:
#> ! `estimate` is required for class metrics but was not provided.
#> ℹ In a metric set, the `estimate` argument must be named.
#> ℹ Example: `my_metrics(data, truth, estimate = my_column)`
```

PR #606, opened 2026-01-21
Add `fall_out()` and `miss_rate()` metrics

### LINKED ISSUE #336: [Feature request] Adding false positive rate (and other basic metrics) as metrics
Not sure if I'm missing something obvious, but I didn't find the false positive rate among the available metrics in yardstick. I know I could simply calculate it by doing `1 - specificity`, but it feels like an unnecessary extra step when there are many other metrics available by default in yardstick.

https://yardstick.tidymodels.org/articles/metric-types.html

EmilHvitfeldt:
To close #336

### DIFF
--- a/R/class-miss_rate.R
+++ b/R/class-miss_rate.R
@@ -32,13 +32,6 @@
 #' that all actual positives were correctly predicted as positive (no false
 #' negatives).
 #'
-#' @author Max Kuhn
-#'
-#' @references
-#'
-#' Fawcett, T. (2006). "An introduction to ROC analysis". *Pattern
-#' Recognition Letters*, 27(8), 861-874.
-#'
 #' @template examples-class
 #'
 #' @export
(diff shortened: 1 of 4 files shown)


R/class-miss_rate.R
```diff
@@ -0,0 +1,256 @@
...
+#' measurement system compared to a reference result (the "truth" or gold
+#' standard). Miss rate is defined as `1 - sensitivity`, or equivalently,
+#' the proportion of positives that are incorrectly classified as negatives.
+#'
+#' Miss rate is also known as the false negative rate (FNR) or the probability
+#' of miss.
+#'
+#' When the denominator of the calculation is `0`, miss rate is undefined.
+#' This happens when both `# true_positive = 0` and `# false_negative = 0`
+#' are true, which mean that there were no true events. When computing binary
```
hfrick:
```suggestion
#' are true, which means that there were no events. When computing binary
```

hfrick (approved):
Looks good 👍

